Issue #31, opened 2023-11-01
Mixed content warnings when running the web app


adityamandaleeka:
From @danroth27 

The web app is hosted on an HTTPS address, but it displays images that are retrieved over plain HTTP, which results in lots of "Mixed content" warnings in the browser dev console.

PR #74, opened 2023-11-13
Serve product images from WebApp, so CatalogApi doesn't have to be publicly reachable

SteveSandersonMS:
Fixes https://github.com/dotnet/eShop/issues/31

This will require more collaboration:

 * ... with @Eilon otherwise it will break the `HybridApp` project. That app is going to need to start knowing the URL to `WebApp` so it can fetch product images from there, *or* the `Mobile.Bff.Shopping` project could be changed to override the `PictureUri` values on products so it points to `WebApp`, using the URL pattern `/product-images/{id}`.
    * The reason it would break is that `HybridApp` shares the `CatalogListItem.razor` component. I've introduced a new interface, `IProductImageUrlProvider`, for which `HybridApp` can provide an implementation that returns URLs on `WebApp`.
    * `WebApp` has an implementation of that interface, but it doesn't need to know absolute URLs because the whole point here is that on the web, you're getting the images from the same origin.
 * ... with the MAUI project (@jamesmontemagno ?). How does it access the product images currently? If it's trying to fetch them directly from `item.PictureUri`, that will be incorrect because we need to stop assuming that CatalogApi is publicly reachable. Most likely it should be changed to fetch the product images from `WebApp` instead, using the URL pattern `/product-images/{id}`.

Given this complexity, I'm not sure we should do this at the last minute, though I do agree it's a change we should make when possible.

### DIFF
--- a/src/WebApp/Extensions/Extensions.cs
+++ b/src/WebApp/Extensions/Extensions.cs
@@ -7,14 +7,14 @@ using Microsoft.IdentityModel.JsonWebTokens;
 
 public static class Extensions
 {
-    private const string ProductImageProxyHttpClient = "product-image-proxy";
-
     public static void AddApplicationServices(this IHostApplicationBuilder builder)
     {
         builder.AddAuthenticationServices();
 
         builder.AddRabbitMqEventBus("EventBus")
                .AddEventBusSubscriptions();
+
+        builder.Services.AddHttpForwarderWithServiceDiscovery();
 
         // Application services
         builder.Services.AddScoped<BasketState>();
@@ -32,8 +32,6 @@ public static class Extensions
 
         builder.Services.AddHttpClient<OrderingService>(o => o.BaseAddress = new("http://ordering-api"))
             .AddAuthToken();
-
-        builder.Services.AddHttpClient(ProductImageProxyHttpClient, o => o.BaseAddress = new("http://catalog-api"));
     }
 
     public static void AddEventBusSubscriptions(this IEventBusBuilder eventBus)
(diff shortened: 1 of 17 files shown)


danmoseley:
as this was done from a private fork, I believe you now need to delete that fork (keep clone locally!) then re-fork now this repo is public, then close this PR and create a new one from your new public fork.

adityamandaleeka:
@danmoseley It doesn't look like this was done from a private fork to me? Or did something change after you posted that comment?

src/WebApp/Extensions/Extensions.cs
```diff
@@ -95,4 +100,26 @@ public static void AddAuthenticationServices(this IHostApplicationBuilder builde
         var user = authState.User;
         return user.FindFirst("name")?.Value;
     }
+
+    public static void MapProductImageProxy(this IEndpointRouteBuilder app)
+    {
+        app.MapGet("/product-images/{id:int}", async (IHttpClientFactory httpClientFactory, int id) =>
+        {
+            var httpClient = httpClientFactory.CreateClient(ProductImageProxyHttpClient);
+            var imageResponse = await httpClient.GetAsync($"/api/v1/catalog/items/{id}/pic");
```
eerhardt, line 109: Can/should Yarp be used for this? That's what "eshopLite" does: https://github.com/dotnet/aspire/blob/cb54d832ce8328c2ac681b031a8c0e0d0f709a7e/samples/eShopLite/MyFrontend/Program.cs#L31

davidfowl: Yep

SteveSandersonMS: Oh that's very nice. I started by trying to do this with Yarp but didn't discover the `MapForwarder` method, and got lost in the weeds with `LoadFromMemory`, trying to call `PathSet` and all kinds of mess. I did make it work in the end but then it was super unclear how to wire it up to service discovery, so I thought it was better not to introduce all that for this sample.

I'll have another go with `MapForwarder`. Is there an example of using this with service discovery somewhere? It needs to be able to make an outbound request to `http://catalogapi` so it needs to use an `HttpClient` that knows about this.

danmoseley:
Ah I saw stevesa: but it's stevesa/

src/WebApp/Extensions/Extensions.cs
eerhardt: > Is there an example of using this with service discovery somewhere? 

Yep! The same app uses the `Microsoft.Extensions.ServiceDiscovery.Yarp` package as well. See https://github.com/dotnet/aspire/blob/ab931f067195a7af0468c18974b864c51a2963cd/samples/eShopLite/MyFrontend/Program.cs#L9

SteveSandersonMS: Excellent, thanks. It works perfectly - this PR now uses that.

SteveSandersonMS:
This is now complete, in that WebApp and HybridApp are both updated to get the images via proxies instead of having the client reference URLs on catalog-api, so I'll merge.

I don't know whether the MAUI app needs a similar change to avoid referencing catalog-api directly from the client but if so that can be handled separately (cc @jamesmontemagno).

SteveSandersonMS:
Looks like merging now requires an approval. @Eilon can you approve this, since you've already been through all this code and wrote part of it?

Eilon (approved):

PR #81, opened 2023-11-14
Add PR validation GitHub Action

captainsafia:


.github/workflows/pr-validation.yml
```diff
@@ -0,0 +1,27 @@
...
+
+
+jobs:  
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup .NET 8
+        uses: actions/setup-dotnet@v3
```
danmoseley, line 21: how does this know to pick 8?

danmoseley: it's fine, was just curious

Directory.Packages.props
```diff
@@ -33,7 +33,7 @@
     <PackageVersion Include="Microsoft.AspNetCore.TestHost" Version="$(AspnetVersion)" />
     <PackageVersion Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="$(AspnetVersion)" />
     <PackageVersion Include="Microsoft.Extensions.Identity.Stores" Version="$(AspnetVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Http.Resilience" Version="8.0.0-rtm.23524.3" />
+    <PackageVersion Include="Microsoft.Extensions.Http.Resilience" Version="8.0.0" />
```
danmoseley, line 36: nit, extract `<MicrosoftExtensionsVersion>` maybe?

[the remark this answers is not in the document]
captainsafia: I tried to avoid making too many version upgrades here to avoid rocking the boat but this one seems safe to do. There are other places where we reference RC packages but we should look at those more closely.

PR #86, opened 2023-11-14
remove internal nuget package sources

sirduke:


danmoseley:
we just did this -- thanks @sirduke, you're quick! 

(would you like to offer a PR for something else?)

PR #87, opened 2023-11-14
markdown lint

danmoseley:
since I did it for dotnet/aspire.


Mishatopkek (approved):

src/Identity.API/README.md
```diff
@@ -1,14 +0,0 @@
-# Containerized eShop - Identity Service
```
danmoseley, line 1: deleted these readmes as they seemed legacy/low value. we can add new ones for each project later if we want.

adityamandaleeka (approved):

PR #96, opened 2023-11-16
Retrieve HybridApp images from web app

Eilon:
@SteveSandersonMS - here's a PR to your PR to have the .NET MAUI Hybrid App retrieve images from the web app.

I considered two approaches:

1. The approach in this PR, which is to retrieve images directly from the HTML to the web app. This is quite simple (1 line of code)
1. An alternative would be to do something more akin to the proxy pattern in the webapp, which is to have the Hybrid App get the images from the catalog-api service (via the mobile-bff), and then generate a `data:base64` URL with the image data. This isn't hard to do, but it's more complicated, and more work.

I think both options are technically correct, but I don't know which way we'd want to lean in this app. If (1) is OK, then we can merge this PR into your PR branch. If we want to do (2), then I can get that done instead.

### DIFF
--- a/src/HybridApp/Services/ProductImageUrlProvider.cs
+++ b/src/HybridApp/Services/ProductImageUrlProvider.cs
@@ -1,0 +1,9 @@
+﻿using eShop.WebAppComponents.Services;
+
+namespace eShop.HybridApp.Services;
+
+public class ProductImageUrlProvider : IProductImageUrlProvider
+{
+    public string GetProductImageUrl(int productId) 
+        => $"{MauiProgram.ProductImageHostUrl}/product-images/{productId}";
+}
--- a/src/HybridApp/MauiProgram.cs
+++ b/src/HybridApp/MauiProgram.cs
@@ -1,4 +1,5 @@
-﻿using eShop.WebAppComponents.Services;
+﻿using eShop.HybridApp.Services;
+using eShop.WebAppComponents.Services;
 using Microsoft.Extensions.Logging;
 
 namespace eShop.HybridApp;
@@ -7,6 +8,8 @@ public static class MauiProgram
 {
     // NOTE: Must have a trailing slash to ensure the full BaseAddress URL is used to resolve relative URLs
     private static string MobileBffBaseUrl = "http://localhost:61632/catalog-api/";
+
+    internal static string ProductImageHostUrl = "https://localhost:7298";
 
     public static MauiApp CreateMauiApp()
     {
@@ -26,6 +29,7 @@ public static class MauiProgram
 #endif
 
         builder.Services.AddHttpClient<CatalogService>(o => o.BaseAddress = new(MobileBffBaseUrl));
+        builder.Services.AddSingleton<IProductImageUrlProvider, ProductImageUrlProvider>();
 
         return builder.Build();
     }


SteveSandersonMS:
Any approach is fine with me. It's a matter of what you're willing to hardcode into HybridApp. It doesn't look like there's currently any way even to configure a different value for `MobileBffBaseUrl`, which would make it difficult to deploy this.

The second option you've listed sounds a little more flexible, though I wonder if it could be simplified a step further. Instead of using any `data:` URLs, could Mobile BFF itself reverse-proxy to catalog-api? That is, Mobile BFF would itself expose endpoints like `/product-images/{id}`, and we already know that the hybrid app is able to reach Mobile BFF via HTTP so that should always be valid. The actual reverse proxying from Mobiel BFF to catalog-api is a trivial one-liner if done in the same way that WebApp does.

Eilon:
> Any approach is fine with me. It's a matter of what you're willing to hardcode into HybridApp. It doesn't look like there's currently any way even to configure a different value for `MobileBffBaseUrl`, which would make it difficult to deploy this.
> 
> The second option you've listed sounds a little more flexible, though I wonder if it could be simplified a step further. Instead of using any `data:` URLs, could Mobile BFF itself reverse-proxy to catalog-api? That is, Mobile BFF would itself expose endpoints like `/product-images/{id}`, and we already know that the hybrid app is able to reach Mobile BFF via HTTP so that should always be valid. The actual reverse proxying from Mobile BFF to catalog-api is a trivial one-liner if done in the same way that WebApp does. Maybe this is exactly what you were proposing anyway, just without the `data:` URLs.

Actually that's entirely different from what I proposed. And a waaay better idea! 😁 I'll give that a shot with the reverse proxy!

(My data URL idea was to download the image into the app (in-memory), base64 encode it, and cram it into a data URL. It would totally work, but is decidedly inefficient.)

src/HybridApp/Services/ProductImageUrlProvider.cs
```diff
@@ -0,0 +1,9 @@
+﻿using eShop.WebAppComponents.Services;
+
+namespace eShop.HybridApp.Services;
+
+public class ProductImageUrlProvider : IProductImageUrlProvider
+{
+    public string GetProductImageUrl(int productId) 
+        => $"{MauiProgram.MobileBffCatalogBaseUrl}api/v1/catalog/items/{productId}/pic";
```
Eilon, line 8: This was even easier than I thought: the Mobile BFF already proxies the entire catalog API, so I just had to use the correct URL, and the image API was already available.

SteveSandersonMS: > This was even easier than I thought: the Mobile BFF already proxies the entire catalog API, so I just had to use the correct URL, and the image API was already available.

Surely it should not do that. The point of Mobile BFF is to publicly expose some endpoints that would not normally be publicly reachable from the internet. If it proxies everything, there's no point even having a Mobile BFF and the app owner might as well set up a public HTTP ingress for CatalogAPI.

Eilon: I think right now it seems a bit odd because the Hybrid App is a read-only view of _only_ the catalog and is not yet a full experience. In the future it would likely need to proxy multiple different services (cart, orders, etc.), plus also have some bespoke mobile-specific APIs (maybe tracking some app preferences/settings, usage insights/telemetry, etc.). But the whole HybridApp is really just 2 pages now (catalog list + item details), so the pattern isn't apparent.

It's probably reasonable to argue that instead it should not proxy _any_ entire service, and instead only specific APIs the app needs.

If you or anyone has thoughts on this I'd be happy to adjust to the pattern. I'm not quite sure what principles we are applying here, so I'd like to follow whatever we recommend.

Eilon: I'll give this a try. Given how simple the app is, it's _probably_ not that hard to do.

Eilon: Alright, it wasn't entirely trivial, but also not *too* complicated. It's a bit ugly, though. The challenge is that I had to add a specific mapping for each catalog-api that the mobile-bff needed to expose to the CatalogService (which is shared between WebApp and HybridApp). I kept the exact same API pattern that catalog-api already had because it means we can share the same CatalogService implementation. That's what a bit yucky: the way we have CatalogService right now *requires* that the exact same catalog APIs are exposed in exactly the same way (except the URL base address). Otherwise we'd need an `ICatalogService` and two different implementations, which would also be a bit yucky.

src/Mobile.Bff.Shopping/Extensions/Extensions.cs
```diff
@@ -2,10 +2,7 @@
 {
     public static void AddApplicationServices(this IHostApplicationBuilder builder)
     {
-        builder.Services
-            .AddReverseProxy()
```
Eilon, line 6: This change removes the free-for-all reverse proxy and instead maps only specific URL patterns (see `Program.cs` below), and forwards those as needed.

src/HybridApp/Services/ProductImageUrlProvider.cs
SteveSandersonMS: That's great - this looks very good now! I don't even think it's ugly TBH - it feels correct to specify each endpoint that you want to expose publicly.

SteveSandersonMS (approved):

src/HybridApp/Services/ProductImageUrlProvider.cs
Eilon: Beauty is in the eye of the BFF.

Eilon:
@SteveSandersonMS - because this is a PR to your PR, please merge at _your_ convenience so I don't accidentally trounce any work you might have pending.

PR #99, opened 2023-11-17
Updating MediatR and switching to keyed services

jbogard:
Fixes this issue called out by @davidfowl : https://x.com/davidfowl/status/1725559031484555616?s=20

Pretty neat way of approaching the design, too.

### DIFF
--- a/src/EventBus/Extensions/EventBusBuilderExtensions.cs
+++ b/src/EventBus/Extensions/EventBusBuilderExtensions.cs
@@ -25,34 +25,7 @@ public static class EventBusBuilderExtensions
         // the consumer can use IKeyedServiceProvider.GetKeyedService<IIntegrationEventHandler>(typeof(T)) to get all
         // handlers for the event type.
 
-        // MediatR assembly scanning breaks with keyed services see https://github.com/jbogard/MediatR/issues/942.
-        // eventBusBuilder.Services.AddKeyedTransient<IIntegrationEventHandler, TH>(typeof(T));
-
-        // Instead, we'll register the handler as transient and use the HandlerTypes on SubscriptionInfo to get all
-        // handlers for the event type.
-        eventBusBuilder.Services.AddTransient<TH>();
-
-        eventBusBuilder.Services.Configure<EventBusSubscriptionInfo>(o =>
-        {
-            // Keep track of all registered event types and their name mapping. We send these event types over the message bus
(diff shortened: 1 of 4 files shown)


src/EventBus/Extensions/EventBusBuilderExtensions.cs
```diff
@@ -25,34 +25,7 @@ public static IEventBusBuilder ConfigureJsonOptions(this IEventBusBuilder eventB
...
         // handlers for the event type.
 
-        // MediatR assembly scanning breaks with keyed services see https://github.com/jbogard/MediatR/issues/942.
-        // eventBusBuilder.Services.AddKeyedTransient<IIntegrationEventHandler, TH>(typeof(T));
-
-        // Instead, we'll register the handler as transient and use the HandlerTypes on SubscriptionInfo to get all
-        // handlers for the event type.
-        eventBusBuilder.Services.AddTransient<TH>();
-
-        eventBusBuilder.Services.Configure<EventBusSubscriptionInfo>(o =>
```
davidfowl, line 35: We still need some of this, but only the list of event types.

This part:

o.EventTypes[typeof(T).Name] = typeof(T);

jbogard: Ah right. Just don't need the `HandlerTypes` business anymore.

src/EventBusRabbitMQ/RabbitMQEventBus.cs
```diff
@@ -200,16 +200,8 @@ private async Task ProcessEvent(string eventName, string message)
         // REVIEW: This could be done in parallel
 
         // Get all the handlers using the event type as the key
-        // foreach (var handler in scope.ServiceProvider.GetKeyedServices<IIntegrationEventHandler>(eventType))
-
-        _subscriptionInfo.HandlerTypes.TryGetValue(eventType, out var handlerTypes);
```
davidfowl, line 205: Need this undeleted as well.

jbogard: Does the keyed services not replace the whole `HandlerTypes` business? That you don't need to keep track of all the handler types to resolve, you rely on the container instead.

davidfowl: Oh you're right here.

davidfowl (approved):

davidfowl:
Thanks @jbogard !

PR #106, opened 2023-11-23
vector type max length is 1024 in db

BiuBiuBiuBiuBooom:
pgsql version : PostgreSQL 14.0 (Debian 14.0-1.pgdg110+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit

pgsql docker image: ankane/pgvector:latest

docker run command:    docker run --hostname=393104f8245b --mac-address=02:42:ac:11:00:03 --env=POSTGRES_HOST_AUTH_METHOD=trust --env=POSTGRES_PASSWORD=e19936df92ee4c1d9fbf10ef139172e3 --env=PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/lib/postgresql/14/bin --env=GOSU_VERSION=1.12 --env=LANG=en_US.utf8 --env=PG_MAJOR=14 --env=PG_VERSION=14.0-1.pgdg110+1 --env=PGDATA=/var/lib/postgresql/data --volume=/var/lib/postgresql/data -p 127.0.0.1:5432 --restart=no --runtime=runc -d ankane/pgvector:latest

screenshot of fail in catalog-api service I encountered: 
Screenshot 2023-11-24 023231
then i did it in my databasee ide to the same pgsql instance, it gave me the same error
Screenshot 2023-11-24 022259

### DIFF
--- a/src/Catalog.API/Infrastructure/EntityConfigurations/CatalogItemEntityTypeConfiguration.cs
+++ b/src/Catalog.API/Infrastructure/EntityConfigurations/CatalogItemEntityTypeConfiguration.cs
@@ -13,7 +13,7 @@ class CatalogItemEntityTypeConfiguration
         builder.Ignore(ci => ci.PictureUri);
 
         builder.Property(ci => ci.Embedding)
-            .HasColumnType("vector(1536)");
+            .HasColumnType("vector(1024)");
 
         builder.HasOne(ci => ci.CatalogBrand)
             .WithMany();
--- a/src/Catalog.API/Infrastructure/Migrations/20231009153249_Initial.Designer.cs
+++ b/src/Catalog.API/Infrastructure/Migrations/20231009153249_Initial.Designer.cs
@@ -74,7 +74,7 @@ namespace eShop.Catalog.API.Infrastructure.Migrations
                         .HasColumnType("text");
 
                     b.Property<Vector>("Embedding")
-                        .HasColumnType("vector(1536)");
+                        .HasColumnType("vector(1024)");
 
                     b.Property<int>("MaxStockThreshold")
                         .HasColumnType("integer");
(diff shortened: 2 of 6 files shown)


BiuBiuBiuBiuBooom:
the error happened at the every beginning, when catalog api service try to create 'Catalog' table

halilkocaoz:
editing [this line](https://github.com/dotnet/eShop/blob/main/src/Catalog.API/Infrastructure/EntityConfigurations/CatalogItemEntityTypeConfiguration.cs#L16) and after then taking migration might be better

BiuBiuBiuBiuBooom:
@dotnet-policy-service agree

BiuBiuBiuBiuBooom:
> editing [this line](https://github.com/dotnet/eShop/blob/main/src/Catalog.API/Infrastructure/EntityConfigurations/CatalogItemEntityTypeConfiguration.cs#L16) and after then taking migration might be better

have committed the missing part

captainsafia:
@roji @stephentoub Is it intentional that we use a vector of unsupported length for storing our embeddings here? It seems off but the value we currently use seems too specific to be unintentional.

stephentoub:
1536 is the length of the embedding vectors created by text-embedding-ada-002 model.

I'm not sure where the cited 1024 maximum is coming from; I've not seen that.

adityamandaleeka:
@BiuBiuBiuBiuBooom This looks like an issue in pgvector that was fixed: https://github.com/pgvector/pgvector/issues/340

BiuBiuBiuBiuBooom:
@adityamandaleeka  sorry guys, u r right. I ever set an alibaba image registry in my docker desktop. when I run the docker pull ankane/pgvector:latest command. it actual gives me an old version of image which is built at least 2 years ago and its version actually is v0.2.0. 

It works just find with a vpn after I remove that registry in my machine. 😭

stephentoub:
Thanks for confirming it works.

PR #113, opened 2023-11-30
Add functional test for Create Order Draft happy path

jflaga:
- also added AutoFixture


captainsafia (commented):
Thanks for submitting this PR!

Is there a chance we can rewrite the test to avoid introducing the new AutoFixture dependency?

I appreciate the convenience it provides when doing test setup but I don't think we should add it for the sake of a single unit test.

jflaga:
Hi @captainsafia .

Thanks for you comment.

AutoFixture is now removed.

captainsafia (approved):

captainsafia:
@jflaga Approved and merged. Thanks for the contribution!

Issue #23, opened 2023-11-01
Remove DDD boilerplate


adityamandaleeka:
From @bricelam:

It is the professional opinion of the EF team that the DDD patterns employed in the Ordering.API and related projects are of no value (and possibly even result in negative value) and can safely be removed for the sake of maintainability.

This includes, but is not limited to:

- [ ] Removing the types under the [src/Ordering.Domain/SeedWork/](https://github.com/dotnet/eShop/tree/main/src/Ordering.Domain/SeedWork) directory
- [ ] Use actual CLR enums for [CardType](https://github.com/dotnet/eShop/blob/main/src/Ordering.Domain/AggregatesModel/BuyerAggregate/CardType.cs) and [OrderStatus](https://github.com/dotnet/eShop/blob/main/src/Ordering.Domain/AggregatesModel/OrderAggregate/OrderStatus.cs)
- [ ] Update [OrderQueries](https://github.com/dotnet/eShop/blob/main/src/Ordering.API/Application/Queries/OrderQueries.cs) to use EF LINQ queries
  - Make ViewModel classes more idiomatic
- [ ] Consider Using `IdentityGuid` as the primary key (instead of the current surrogate key) for [Buyer](https://github.com/dotnet/eShop/blob/main/src/Ordering.Domain/AggregatesModel/BuyerAggregate/Buyer.cs)
- [ ] Review the unusual use of transactions and the SaveChanges call in [TransactionBehavior](https://github.com/dotnet/eShop/blob/main/src/Ordering.API/Application/Behaviors/TransactionBehavior.cs)

adityamandaleeka:
@SamMonoRT is this on your team's radar? Any idea when this will be worked on?

SamMonoRT:
No wasn't on my radar, but now is. I'll review and update here, but at the moment this is below our cut line for 9.0 and we may not get to it anytime soon, given other priorities.

PR #115, opened 2023-12-01
Update OrderQueries to use linq queries

AlmarAubel:
- Changed all plain SQL queries in `OrderQueries` to LINQ queries.
- Made `BuyerId` an auto-property to enable its use in LINQ queries.
- Used DbContext to query card types. Uncertain whether it's better to create a new repository or add to an existing one.

This addresses one of the points mentioned by @adityamandaleeka in issue #23.


AlmarAubel:
@dotnet-policy-service agree

adityamandaleeka:
Tagging @roji @ajcvickers for best practices review here.

Would be great if we could see a before/after perf comparison for this sort of thing.

AlmarAubel:
I've completed the benchmarking as requested to assess the performance implications of transitioning from raw SQL queries to EF Core, followed by further optimizations. Initially, the queries were naively translated to EF Core with minimal modifications to the classes used in querying. Subsequently, I implemented optimizations, primarily modifying certain getter methods with a private backing field to properties with a private setter. This adjustment aids EF Core in better optimizing queries, especially for expressions like 'total = (double)ob.order.GetTotal()'.

## Results
Here are the benchmark results:

| Method                | Mean       | Error     | StdDev    | Ratio  | RatioSD | Allocated  | Alloc Ratio |
|---------------------- |-----------:|----------:|----------:|-------:|--------:|-----------:|------------:|
| RawSql                |   1.257 ms | 0.0248 ms | 0.0194 ms |   1.00 |    0.00 |    9.55 KB |        1.00 |
| Naive                 | 136.107 ms | 2.1718 ms | 0.1190 ms | 108.18 |    0.93 | 2900.74 KB |      303.78 |
| NaiveAsNoTracking     | 134.255 ms | 1.3476 ms | 0.8914 ms | 106.77 |    1.93 | 3103.67 KB |      325.03 |
| Optimized             |   1.143 ms | 0.0199 ms | 0.0176 ms |   0.91 |    0.02 |   23.25 KB |        2.44 |
| OptimizedAsNoTracking |   1.152 ms | 0.0286 ms | 0.0330 ms |   0.92 |    0.03 |   26.04 KB |        2.73 |


**Analysis:**

- **Raw SQL to Naive EF Core**: The transition from raw SQL to a naive EF Core implementation resulted in a significant increase in execution time and memory allocation. This is expected as the naive approach does not take advantage of EF Core's optimization capabilities.
  
- **Naive to Optimized EF Core**: Post optimization, there's a dramatic improvement, bringing the performance almost on par with the raw SQL, both in terms of execution time and memory usage. The optimizations primarily involved restructuring getter methods for better EF Core query translation.

- **NoTracking Variants**: The 'AsNoTracking' variants show similar performance to their tracking counterparts, with a slight increase in memory usage. This indicates that tracking does not significantly impact performance in this context.

This comparison clearly illustrates the importance of proper query structuring and optimization in EF Core to achieve performance close to raw SQL. I believe these optimizations represent a more effective approach for this scenario.

## SQL 
For reference, here are the queries:

### Raw SQL: 
```sql     
SELECT o."Id" AS ordernumber, o."OrderDate" AS date, os."Name" AS status, SUM(oi."Units" * oi."UnitPrice") AS total
   FROM ordering.orders AS o
   LEFT JOIN ordering."orderItems" AS oi ON o."Id" = oi."OrderId"
   LEFT JOIN ordering.orderstatus AS os ON o."OrderStatusId" = os."Id"
   LEFT JOIN ordering.buyers AS ob ON o."BuyerId" = ob."Id"
   WHERE ob."IdentityGuid" = @userId
   GROUP BY o."Id", o."OrderDate", os."Name"
   ORDER BY o."Id"
```

###  Optimized: 
```sql     
SELECT o."Id" AS ordernumber, o."OrderDate" AS date, o0."Name" AS status, (
    SELECT COALESCE(sum(o1."UnitPrice" * o1."Units"::numeric), 0.0)
    FROM ordering."orderItems" AS o1
    WHERE o."Id" = o1."OrderId")::double precision AS total
FROM ordering.orders AS o
         INNER JOIN ordering.buyers AS b ON o."BuyerId" = b."Id"
         INNER JOIN ordering.orderstatus AS o0 ON o."OrderStatusId" = o0."Id"
WHERE b."IdentityGuid" = @___buyerId_0
```
### Naive: 
```sql     
SELECT o."Id", o."BuyerId", o."Description", o."OrderDate", o."OrderStatusId", o."PaymentMethodId", o."Address_City", o."Address_Country", o."Address_State", o."Address_Street", o."Address_ZipCode", b."Id", o0."Id", o1."Id", o1."OrderId", o1."ProductId", o1."UnitPrice", o1."Units", o1."Discount", o1."PictureUrl", o1."ProductName", o0."Name", o2."Id", o2."OrderId", o2."ProductId", o2."UnitPrice", o2."Units", o2."Discount", o2."PictureUrl", o2."ProductName"
FROM ordering.orders AS o
         INNER JOIN ordering.buyers AS b ON o."BuyerId" = b."Id"
         INNER JOIN ordering.orderstatus AS o0 ON o."OrderStatusId" = o0."Id"
         LEFT JOIN ordering."orderItems" AS o1 ON o."Id" = o1."OrderId"
         LEFT JOIN ordering."orderItems" AS o2 ON o."Id" = o2."OrderId"
WHERE b."IdentityGuid" = @___buyerId_0
ORDER BY o."Id", b."Id", o0."Id", o1."Id"
```

## Next steps
Based on these insights, I propose further refactoring to enhance our application's performance and maintainability. If your agrees with this approach, I plan to:

1. **Refactor all private fields and getter methods into auto-properties with private setters.** This will further align with the EF Core optimization strategies and potentially improve performance.
2. **Adjust the EF configuration and other related code accordingly.** These changes will ensure our codebase remains consistent and optimized for EF Core usage.

# Disclaimer :) 
Please let me know if you agree with this direction. Upon confirmation, I'll proceed with the necessary modifications and update the PR accordingly.

Finally, I'd like to emphasize that benchmarking is a complex area, often with nuanced results. I've tried to ensure that the methodology I chose is suitable for this context, but I'm open to feedback or alternative approaches that could refine our understanding of the performance impacts.

You can find the code used for the benchmarks [here](https://github.com/AlmarAubel/eShop/blob/add-benchmarks/benchmarks/Ordering.Benchmarks/GetOrdersBenchmark.cs).

AlmarAubel:
The GetOrders async is even faster then the raw-sql. Tomorrow I will share full results of the other query functions.

| Method               | Mean     | Error     | StdDev    | Allocated |
|--------------------- |---------:|----------:|----------:|----------:|
| RawSql_GetOrderAsync | 2.473 ms | 0.3614 ms | 0.4162 ms |  27.39 KB |
| EfCore_GetOrderAsync | 1.326 ms | 0.0432 ms | 0.0480 ms |  18.04 KB |

AlmarAubel:
# Benchmark Comparison of Raw SQL vs EF Core Queries**

I've recently conducted a benchmark to compare the performance of raw SQL implementations against EF Core queries.

Firstly, it's important to note that I excluded the connection setup time from the benchmark. This decision was made to focus solely on the query execution performance. Including connection setup could introduce variability unrelated to the query performance itself, especially in scenarios where connection pooling or other optimizations are in play. By excluding this, we get a clearer picture of how the queries themselves are performing.

Here are the results:

| Method | Mean | Error | StdDev | Allocated |
|-------------------------|-----------:|----------:|----------:|----------:|
| RawSql_GetOrderAsync | 2,005.2 us | 237.79 us | 244.19 us | 27.34 KB |
| EfCore_GetOrderAsync | 1,950.0 us | 246.93 us | 264.21 us | 18.1 KB |
| RawSql_GetOrdersAsync | 1,766.5 us | 363.29 us | 403.79 us | 9.55 KB |
| EfCore_GetOrdersAsync | 1,367.0 us | 154.21 us | 171.40 us | 30 KB |
| RawSql_GetCardTypesAsync | 953.6 us | 84.56 us | 86.84 us | 2.92 KB |
| EfCore_GetCardTypesAsync | 865.8 us | 37.38 us | 41.55 us | 7.46 KB |

The results show that EF Core queries generally perform better in terms of execution time, although they sometimes allocate more memory. This trade-off might be worth considering based on the project's specific needs and constraints.

AlmarAubel:
@adityamandaleeka any change to have a look?

src/Ordering.Domain/AggregatesModel/OrderAggregate/Order.cs
```diff
@@ -195,4 +195,15 @@ public decimal GetTotal()
     {
         return _orderItems.Sum(o => o.GetUnits() * o.GetUnitPrice());
     }
+    
+    
+    public DateTime GetOrderDate()
```
roji, line 200: I'd simply have an OrderDate property (and a BuyerId property). The general feeling in the team is that the patterns used here - e.g. avoiding properties and having private fields with Set/Get methods - is an anti-pattern.

But this can be done in a separate PR. Otherwise this can be an expression-bodied method for nicer C#.

src/Ordering.Domain/AggregatesModel/OrderAggregate/Order.cs
```diff
@@ -102,7 +102,7 @@ public void SetPaymentId(int id)
 
     public void SetBuyerId(int id)
     {
-        _buyerId = id;
+        BuyerId = id;
```
roji, line 105: Especially if BuyerId is exposed as a public property, having a public SetBuyerId doesn't seem to make sense - we can just make the setter public.

(this code was written using various DDD-inspired patterns that the team doesn't feel are a good idea)

src/Ordering.API/Application/Queries/OrderQueries.cs
```diff
@@ -1,85 +1,57 @@
...
-                oi."UnitPrice" AS unitprice, oi."PictureUrl" AS pictureurl
-            FROM ordering.Orders AS o
-            LEFT JOIN ordering."orderItems" AS oi ON o."Id" = oi."OrderId"
-            LEFT JOIN ordering.orderstatus AS os ON o."OrderStatusId" = os."Id"
-            WHERE o."Id" = @id
-            """,
-            new { id });
-
-        if (result.AsList().Count == 0)
+        var order = await orderRepository.GetAsync(id);
```
roji, line 8: GetAsync in OrderRepository is implemented particularly inefficient, doing Find and then two LoadAsync for the related entities - that's 3 roundtrips where only one should be done. Given that the previous code just issued direct SQL - bypassing the repository - maybe it's OK to bypass the repository and just integrate a LINQ query here.

Note that this is what's done below in GetOrdersFromUserAsync; we should be consistent here, either always going through the repository (in which case the LINQ query in GetOrdersFromUserAsync needs to move to the repository), or not doing it (which I'd recommend at least for now).

Stepping back, there's a fundamental question here of whether we want to go use the repository abstraction or not. The general feeling in the team is that a repository abstraction over EF (IOrderRepository) isn't something that should be done by default, and makes sense specifically in the following cases:

* Multiple databases are being targeted, and LINQ queries require tweaking on a per-database basis.
* Unit tests are desired which don't utilize the production database system; the repository layer allows mocking at that point (much better than swapping EF providers with InMemory/Sqlite). However, we don't believe this should be done by default - testing against the real database with e.g. testcontainers is generally a better approach, at least as a baseline (these need to be done anyway to test that queries execute correctly against the real database system).

We should maybe discuss exactly what it is we're trying to show/achieve with eshop, but there's a good chance I'd be for removing the repository layer altogether.

src/Ordering.Infrastructure/EntityConfigurations/OrderEntityTypeConfiguration.cs
```diff
@@ -14,9 +14,9 @@ public void Configure(EntityTypeBuilder<Order> orderConfiguration)
         //Address value object persisted as owned entity type supported since EF Core 2.0
         orderConfiguration
             .OwnsOne(o => o.Address);
-
+        
         orderConfiguration
-            .Property("_buyerId")
+            .Property(o=> o.BuyerId)
```
roji, line 19:
```suggestion
            .Property(o => o.BuyerId)
```

src/Ordering.Infrastructure/EntityConfigurations/OrderEntityTypeConfiguration.cs
```diff
@@ -14,9 +14,9 @@ public void Configure(EntityTypeBuilder<Order> orderConfiguration)
         //Address value object persisted as owned entity type supported since EF Core 2.0
         orderConfiguration
             .OwnsOne(o => o.Address);
-
+
```
roji, line 17:
```suggestion

```

src/Ordering.Infrastructure/EntityConfigurations/OrderEntityTypeConfiguration.cs
```diff
@@ -40,7 +40,7 @@ public void Configure(EntityTypeBuilder<Order> orderConfiguration)
 
         orderConfiguration.HasOne<Buyer>()
             .WithMany()
-            .HasForeignKey("_buyerId");
+            .HasForeignKey(o=> o.BuyerId);
```
roji, line 43:
```suggestion
            .HasForeignKey(o => o.BuyerId);
```

src/Ordering.API/Application/Queries/OrderQueries.cs
```diff
@@ -1,85 +1,57 @@
...
-            total = 0
-        };
-
-        foreach (dynamic item in result)
-        {
-            var orderitem = new Orderitem
+        return await context.Orders
+            .Include(o => o.OrderItems)
+            .Include(o => o.OrderStatus)
+            .Join(context.Buyers,
```
roji, line 40: Given that Order has a BuyerId (foreign key), there should be a regular relationship between Orders and Buyers, and this should be a simple Include... I'm really not sure why it wasn't modelled this way.

But this seems OK as a temporary thing.

src/Ordering.API/Application/Queries/OrderQueries.cs
```diff
@@ -1,85 +1,57 @@
...
+                ordernumber = ob.order.Id,
+                date = ob.order.GetOrderDate(),
+                status = ob.order.OrderStatus.Name,
+                total = (double)ob.order.GetTotal()
+            })
+            .ToListAsync();
     }
+
+    public async Task<IEnumerable<CardType>> GetCardTypesAsync() => 
+        await context.CardTypes.Select(c=> new CardType { Id = c.Id, Name = c.Name }).ToListAsync();
```
roji, line 56: Why the projection to Id and Name only? Should this just return the CareTypes as they are?

roji (commented):
This generally looks OK - I think it makes sense to use LINQ here wherever we can, rather than SQL.

But aside from that, there's a whole lot that probably needs to be cleaned up/refactored here - the design is generally very over-engineered, with layers and an unneeded repository abstraction, etc.

src/Ordering.Domain/AggregatesModel/OrderAggregate/Order.cs
AlmarAubel: Great to hear that, I was also not a fan of using methods for properties. However, I was perhaps overly cautious about rewriting everything since I wasn't sure if there was another reason behind this approach.

src/Ordering.API/Application/Queries/OrderQueries.cs
AlmarAubel: Yes could be but in the context we use  eShop.Ordering.Domain.AggregatesModel.BuyerAggregate.CardType and this query returns  eShop.Ordering.API.Application.Queries.CardType. So that would change the public API of this query class.

src/Ordering.API/Application/Queries/OrderQueries.cs
AlmarAubel: Thank you for your feedback. In my benchmarks, using the repository as the "naive approach" indeed proved to be quite slow. However, in my second branch where I integrated benchmarks, abandoning the repository led to significantly improved performance, almost matching that of raw SQL queries. So great work by the EF Core team!

For query operations, bypassing the repository seems to be a viable and efficient approach.

Additionally, I'd like to highlight another potential use case for the repository pattern. This involves retrieving aggregate roots from the database, ensuring that they are consistently and correctly fetched. Where this cannot be achieved with EF (configuration) This consistency is crucial, especially in complex domain models where an aggregate root encompasses various related entities, maintaining their invariants and business rules.

I apologize for any confusion caused. I've already addressed some issues on the benchmark branch and was awaiting further confirmation. Initially, I was hesitant to refactor extensively, but after reviewing your comments, it's clear that our perspectives align.

Moving forward, I'm inclined to consider removing the repository layer, especially given the current context and objectives with the eshop project.

roji: > This involves retrieving aggregate roots from the database, ensuring that they are consistently and correctly fetched.

That's interesting.. If the intent is to always fetch the related entities (which is how aggregate root is sometimes interpreted), then it probably makes more sense to model this as owned entities, or to use the auto-include feature; that ensures that whenever a principal is loaded, its dependents are loaded as well. Though it's indeed possible that when this code was written these two features weren't yet available.

In any case, I'm not sure eshop should necessarily be showcasing specific modeling patterns (that's a genuine open question - the exact purpose of this sample app is't necessarily 100% defined). I'd lean towards doing the simpler thing here rather than something very involved with patterns that may not be necessary (or even helpful) for most cases. 

> I apologize for any confusion caused.

No need to apologize at all! This is bringing the code forward. If you intend to continue working on this, I indeed recommend doing step-by-step changes for easier review/discussion rather than a big refactor all at once.

> It seems that the use of LoadAsync in the repository was implemented before the availability of AsSplitQuery

That's definitely possible - though I doubt that the performance issues of single query (e.g. cartesian explosion due to joins) was the reason this was written this way...

src/Ordering.API/Application/Queries/OrderQueries.cs
roji: I see, OK.

src/Ordering.Domain/AggregatesModel/OrderAggregate/Order.cs
roji: I'm honestly not sure where this comes from - there's the comment attributing this to "DDD best practices", but I'm not sure exactly what that refers to. In any case, it's fine to leave things as-is and clean this up in a separate pass later.

roji:
@AlmarAubel let me know when you're done with adjustments here by requesting another review.

src/Ordering.API/Application/Queries/OrderQueries.cs
AlmarAubel: Small question can I combine the refactoring of order and orderitem where I will introduce the properties (instead of the getter and setter methods) and the refactoring to LINQ or is it better to make separate PR for that change?

roji: Sure, as long as it's simple stuff (introduce properties instead of get/set methods), no problem.

src/Ordering.API/Application/Queries/OrderQueries.cs
AlmarAubel: I've added a navigation property to the order entity and updated the EF configuration. However, I'm hesitant about this approach due to potential conflicts with Domain-Driven Design (DDD) principles, specifically regarding aggregates. Key concerns include:

1. **Violation of Aggregate Boundaries**: Navigation properties can create tight coupling between aggregates, undermining the autonomy and clear boundaries essential in DDD.

2. **Encouragement of Lazy Loading**: The use of navigation properties often leads to lazy loading, which can result in performance issues.

AlmarAubel:
@roji, I think I've addressed all the issues you raised. Can you review it again?  Wishing you nice Christmas days!

src/Ordering.Infrastructure/EntityConfigurations/OrderItemEntityTypeConfiguration.cs
```diff
@@ -15,23 +15,23 @@ public void Configure(EntityTypeBuilder<OrderItem> orderItemConfiguration)
         orderItemConfiguration.Property<int>("OrderId");
 
         orderItemConfiguration
-            .Property("_discount")
+            .Property(o => o.Discount)
```
roji, line 18: FYI now that these are regular properties, doing HasColumnName is no longer needed and can be removed.

src/Ordering.Domain/AggregatesModel/OrderAggregate/Order.cs
```diff
@@ -193,17 +180,6 @@ private void StatusChangeException(OrderStatus orderStatusToChange)
...
-    
-    public DateTime GetOrderDate()
-    {
-        return _orderDate;
-    }
-    
-    public string GetDescription()
-    {
-        return _description;
+        return _orderItems.Sum(o => o.Units * o.UnitPrice);
```
roji, line 189: nit: can do expression-bodied

AlmarAubel:
@roji  hopefully if fixed all the issues

AlmarAubel:
@adityamandaleeka  or @roji  I fixed the merge conflicts could you please do a final review?

src/Ordering.Infrastructure/EntityConfigurations/OrderItemEntityTypeConfiguration.cs
```diff
@@ -15,23 +15,18 @@ public void Configure(EntityTypeBuilder<OrderItem> orderItemConfiguration)
         orderItemConfiguration.Property<int>("OrderId");
 
         orderItemConfiguration
-            .Property("_discount")
-            .HasColumnName("Discount");
+            .Property(o => o.Discount);
```
roji, line 18: Once you stop configuring an explicit column name, all of these are no longer necessary and can be removed.

In general, it would be good to clean up the EF model configuration to be minimal and only configure what is actually necessary (but we can do further cleanup in subsequent PRs).

roji (changes requested):
This is starting to looking a lot cleaner - thanks for your efforts!

I've pointed out one last simplification we can do after your cleanup, once you do that I'll approve and merge. I can see various additional opporunities for cleanup/simplification in EF code in this application, but we can do that with follow-up PRs later - I hope you're interested in continuing this work!

src/Ordering.Infrastructure/EntityConfigurations/OrderItemEntityTypeConfiguration.cs
AlmarAubel: Yeah no problem but that requires to Change the Order(Item) Aggragates. Implicit configuration will not work on Init only props. In order to make it work i have  added a private setter to all init only props.

AlmarAubel: @roji  pushed the changes

roji (approved):

roji:
@adityamandaleeka this looks ready to merge form my side - but I'm not a maintainer (should I be added?)

src/Ordering.Infrastructure/EntityConfigurations/OrderItemEntityTypeConfiguration.cs
roji: Thanks!

roji:
Thanks @adityamandaleeka!

Issue #116, opened 2023-12-01
Using Records with Primary Constructors instead of Class for Commands / Queries


hanyfakhry:
Great work with all the amazing new features, one enhancement i believe it can be done to move commands from classes to be records for example **CancelOrderCommand**:

```
public class CancelOrderCommand(int orderNumber) : IRequest<bool>
{
    [DataMember]
    public int OrderNumber { get; set; } = orderNumber;
}
```
To Be

`public record CancelOrderCommand(int OrderNumber) : IRequest<bool>;`

This will make the Commands more readable and less code

PR #123, opened 2023-12-08
Using Records with Primary Constructors instead of Class for Commands…

saiphanish:
… / Queries #116 Using Records with Primary Constructors instead of Class for Command

@davidfowl  can you please review my changes. 

This fixes open issue  #116 - Using Records with Primary Constructors instead of Class for Command

### DIFF
--- a/src/Ordering.API/Application/Commands/SetStockRejectedOrderStatusCommand.cs
+++ b/src/Ordering.API/Application/Commands/SetStockRejectedOrderStatusCommand.cs
@@ -1,17 +1,3 @@
 ﻿namespace eShop.Ordering.API.Application.Commands;
 
-public class SetStockRejectedOrderStatusCommand : IRequest<bool>
-{
-
-    [DataMember]
-    public int OrderNumber { get; private set; }
-
-    [DataMember]
-    public List<int> OrderStockItems { get; private set; }
-
-    public SetStockRejectedOrderStatusCommand(int orderNumber, List<int> orderStockItems)
-    {
-        OrderNumber = orderNumber;
-        OrderStockItems = orderStockItems;
-    }
-}
+public record SetStockRejectedOrderStatusCommand(int OrderNumber, List<int> OrderStockItems) : IRequest<bool>;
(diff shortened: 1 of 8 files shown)


src/Ordering.API/Application/Commands/SetStockRejectedOrderStatusCommand.cs
```diff
@@ -1,17 +1,3 @@
 ﻿namespace eShop.Ordering.API.Application.Commands;
 
-public class SetStockRejectedOrderStatusCommand : IRequest<bool>
-{
-
-    [DataMember]
```
captainsafia, line 6: We're losing the `DataMember` attribution on this property. Have we confirmed this doesn't cause any issues?

tests/Ordering.UnitTests/Application/SetStockRejectedOrderStatusCommandTest.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    [Fact]
+    public void Set_Stock_Rejected_OrderStatusCommand_Check_Serialization()
+    {
+        // Arrange
+        var command = new SetStockRejectedOrderStatusCommand(123, new List<int> { 1, 2, 3 });
+
+        // Act
+        var json = JsonConvert.SerializeObject(command);
+        var deserializedCommand = JsonConvert.DeserializeObject<SetStockRejectedOrderStatusCommand>(json);
```
davidfowl, line 15: Why is this test using JSON.NET?

src/Ordering.API/Application/Commands/SetStockRejectedOrderStatusCommand.cs
saiphanish: Yes even though we don't use [DataMember] attribute, by using records, the properties are automatically considered for serialization by default.

Added unit test SetStockRejectedOrderStatusCommandTest to test if serialization is working as expected.

tests/Ordering.UnitTests/Application/SetStockRejectedOrderStatusCommandTest.cs
saiphanish: Updated the code to use System.Text.Json.JsonSerializer instead of JSON.NET.

jamesmontemagno (approved):

captainsafia (approved):
Verified that purchase flow works locally with this change.

Thanks for the contribution!

saiphanish:
Thanks @captainsafia  for merging my changes

PR #125, opened 2023-12-10
Fix editorconfig path in solution

marioleed:
Just a small fix where eShop.sln was having the wrong path to .editorconfig.


captainsafia (approved):

captainsafia:
Thank you for this improvement, @marioleed!

PR #135, opened 2023-12-14
Added catalog functional test Cases.

SrushtiPasari94:
Added functional test cases for catalog functionalities.


tests/Catalog.FunctionalTests/CatalogApiTests.cs
```diff
@@ -87,4 +87,184 @@ public async Task UpdateCatalogItemWorksWithPriceUpdate()
...
+        // Assert   
+        response.EnsureSuccessStatusCode();
+        var body = await response.Content.ReadAsStringAsync();
+        var result = JsonSerializer.Deserialize<List<CatalogBrand>>(body, _jsonSerializerOptions);
+
+        //// Assert       
+        Assert.Equal(13, result.Count);
+        Assert.NotNull(result);
+
+    }
```
captainsafia, line 266: Remove extra whitespace?

tests/Catalog.FunctionalTests/CatalogApiTests.cs
```diff
@@ -87,4 +87,184 @@ public async Task UpdateCatalogItemWorksWithPriceUpdate()
...
+    {
+        // Act
+        var response = await _httpClient.GetAsync("api/v1/catalog/catalogbrands");
+
+        // Assert   
+        response.EnsureSuccessStatusCode();
+        var body = await response.Content.ReadAsStringAsync();
+        var result = JsonSerializer.Deserialize<List<CatalogBrand>>(body, _jsonSerializerOptions);
+
+        //// Assert
```
captainsafia, line 262:
```suggestion
        // Assert       
```

Here and elsewhere.

tests/Catalog.FunctionalTests/CatalogApiTests.cs
captainsafia, line 1: Can we add tests for the `POST` action now that it has been fixed?

I didn't run into any issues with the `DELETE` endpoint either so those can be resolved too.

tests/Catalog.FunctionalTests/CatalogApiTests.cs
```diff
@@ -87,4 +87,184 @@ public async Task UpdateCatalogItemWorksWithPriceUpdate()
...
+
+        //// Assert       
+        Assert.Equal(13, result.Count);
+        Assert.NotNull(result);
+
+    }
+
+
+
 }
```
jamesmontemagno, line 270:
```suggestion
        Assert.NotNull(result);
    }
}
```

tests/Catalog.FunctionalTests/CatalogApiTests.cs
SrushtiPasari94: Removed

tests/Catalog.FunctionalTests/CatalogApiTests.cs
SrushtiPasari94: Added test case for Post and Delete action.

tests/Catalog.FunctionalTests/CatalogApiTests.cs
```diff
@@ -87,4 +87,224 @@ public async Task UpdateCatalogItemWorksWithPriceUpdate()
...
+        var body = await response.Content.ReadAsStringAsync();
+        var result = JsonSerializer.Deserialize<List<CatalogBrand>>(body, _jsonSerializerOptions);
+
+        // Assert       
+        Assert.Equal(13, result.Count);
+        Assert.NotNull(result);
+    }
+
+    [Fact]
+    public async Task AddDeleteCatalogItem()
```
captainsafia, line 267: Can we split these up into two tests? One for adding and one for deleting.

SrushtiPasari94: I can do that. 
The reason of testing in same test case is:  If adding to the particular catalog item(which is getting deleted in delete test) in post method will fail then Delete first assert statement will fails as it will return "NotFound" instead of "NoContent".
Also, if we run delete test case before post test case it will fail as I am deleting the catalog item added in Post test.
Considering that in mind I wrote both functionalities testing together.

captainsafia: Yeah, I'm fine writing one test case that exclusively tests post and then another that tests delete. There's no strong requirement to need to insert a new element in order to delete it since we setup the test DB from for each run.

SrushtiPasari94: @captainsafia Resolving this based on your comment. Thanks

captainsafia (approved):

PR #139, opened 2023-12-19
Upgrade to stable SK 1.0.1

stephentoub:
After this, we should do a little refactoring to actually inject the Kernel via DI, e.g. in central configuration somewhere, we'd AddOpenAIChatCompletion to the service container and AddKernel, and then in the razor component we inject Kernel, rather than building one in ChatState.


captainsafia (approved):
LGTM -- perhaps we should file issues to track the follow-up work?

PR #147, opened 2024-01-04
Inject Kernel into ChatState via DI

### LINKED ISSUE #145: Inject Kernel into ChatState via DI
Today, configuration information is passed into ChatState, which then uses that to construct an OpenAIClient and Kernel.  Now that Kernel is more friendly to DI, and with an OpenAIClient Aspire component in the works, we should change this to have the Kernel injected via DI.  Then ChatState won't need to be concerned about any of the configuration or even which AI service is being used.

eerhardt:
Instead of building up the kernel and inspecting configuration manually in the Chatbot component.

Also allow for an optional Endpoint to be configured, to allow working with Azure OpenAI.

Finally, also adjust the global.json setting to allow for newer SDKs to be used.

Fix #145


src/WebApp/AIOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+public class OpenAIOptions
+{
+    /// <summary>OpenAI API key for accessing embedding LLM.</summary>
+    public string? ApiKey { get; set; }
+
+    /// <summary>Optional endpoint for which OpenAI API to access.</summary>
+    public string? Endpoint { get; set; }
+
+    /// <summary>The name of the chat model to use.</summary>
+    public string ChatName { get; set; } = "gpt-3.5-turbo-16k";
```
eerhardt, line 18: @stephentoub - `ChatName` or `ChatModel`? The existing code used both (ChatModel in code, ChatName in appsettings.json).

src/Catalog.API/AIOptions.cs
```diff
@@ -9,7 +9,10 @@ public class AIOptions
 public class OpenAIOptions
 {
     /// <summary>OpenAI API key for accessing embedding LLM.</summary>
-    public string APIKey { get; set; }
+    public string ApiKey { get; set; }
+
+    /// <summary>Optional endpoint for which OpenAI API to access.</summary>
+    public string Endpoint { get; set; }
```
stephentoub, line 15: I'd previously removed this as an option in order to keep eShop agnostic to Azure. I'm fine if we want to bring it back, but OpenAI won't use this; it's specific to Azure OpenAI.

src/WebApp/AIOptions.cs
stephentoub, line 18: Same comment as above about the endpoint. Also this is the model name in the case of OpenAI, but the deployment name in the case of Azure OpenAI.

src/WebApp/Components/Chatbot/Chatbot.razor
```diff
@@ -53,10 +53,11 @@
 
     protected override async Task OnInitializedAsync()
     {
-        if (ChatConfig.TryReadFromConfig(Configuration, out var chatConfig))
+        var kernel = ServiceProvider.GetService<Kernel>();
```
eerhardt, line 56: @SteveSandersonMS @javiercn - this was the way I figured out how to make an "optional" DI service work in Blazor. See https://github.com/dotnet/razor/issues/7653.

Is there a better way?

src/Catalog.API/AIOptions.cs
eerhardt: Is making it optional good enough? I left it out of the appsetting.json comment so people didn't feel like they were required to supply it.

Also - is there any documentation on how to enable the AI pieces in eshop? I couldn't find any by simple searching.

src/WebApp/AIOptions.cs
eerhardt: I'll add a remark here that needs to be "deployment name" if you are using Azure OpenAI.

src/WebApp/Components/Chatbot/Chatbot.razor
```diff
@@ -8,8 +8,8 @@
 @inject CatalogService CatalogService
 @inject BasketState BasketState
 @inject AuthenticationStateProvider AuthenticationStateProvider
-@inject IConfiguration Configuration
 @inject ILoggerFactory LoggerFactory
+@inject IServiceProvider ServiceProvider
```
stephentoub, line 12: Does razor have any notion of an `@inject` that's optional, i.e. the equivalent of GetService rather than GetRequiredService? If yes, then this could be that with Kernel, and then the code below doesn't need to call GetService and can just do what it does subsequently, checking it for null.

stephentoub (approved):
Thanks.

src/WebApp/AIOptions.cs
stephentoub: If we're just talking about OpenAI, ChatModel is better. If we're also talking about Azure OpenAI, this is either a deployment name for Azure OpenAI or a model name for OpenAI.

src/Catalog.API/AIOptions.cs
stephentoub: > Also - is there any documentation on how to enable the AI pieces in eshop? I couldn't find any by simple searching.

I didn't write any, so probably not?

> Is making it optional good enough?

@adityamandaleeka?

src/WebApp/Components/Chatbot/Chatbot.razor
stephentoub: Nevermind, I see your comment/question above now: https://github.com/dotnet/eShop/pull/147#discussion_r1442177118.

src/Catalog.API/AIOptions.cs
eerhardt: Just some background on why I did this - I already have an Azure account, and so it was easier for me to use Azure OpenAI than to sign up for an OpenAI account. I figured I'd save the next person this trouble in the future.

Are there other deployments of OpenAI (ex. AWS) that this Endpoint would be useful for?

stephentoub: > Are there other deployments of OpenAI (ex. AWS) that this Endpoint would be useful for?

Possibly? I've seen some projects that try to emulate the OpenAI REST surface area in order to let you host a model locally but pretend to be OpenAI, in which case it's possible this would let you use that, e.g. by setting the endpoint to the localhost url where you're hosting the server, e.g.
https://github.com/mudler/LocalAI
https://github.com/THUDM/ChatGLM3
etc.

I've not tried, though.

SteveSandersonMS: > I already have an Azure account, and so it was easier for me to use Azure OpenAI than to sign up for an OpenAI account

Me too - thanks!

BTW if we think plugging in completely different inference backends (e.g., Llama 2 running locally or whatever) then it would be nicer to have a C# interface people can implement rather than making them use the OpenAI REST APIs. I suppose Semantic Kernel is already meant to be an abstraction over different inference mechanisms but I don't know whether we'd need to add further abstraction.

src/WebApp/Components/Chatbot/Chatbot.razor
SteveSandersonMS: > Is there a better way?

Nope, currently this is the way.

SteveSandersonMS: If the `GetService` call bothers you, another fairly idiomatic option would be having some helper that gets the kernel and returns null or throws if the required config is absent. That could be an extension method on `IServiceProvider`. But what you have here is completely reasonable already.

src/Catalog.API/AIOptions.cs
stephentoub: >  I suppose Semantic Kernel is already meant to be an abstraction over different inference mechanisms

Yes. And it's a priority for us to ensure all the right "connectors" are available for SK to make it easy to plug in different models, running both locally and remotely.

src/WebApp/Components/Chatbot/Chatbot.razor
eerhardt: > If the GetService call bothers you

It's mostly injecting `IServiceProvider` that bothers me. It's not a huge concern, but it is usually frowned upon - https://stackoverflow.com/questions/70264338/inject-iserviceprovider-into-an-api-controller-or-individual-services.

> what you have here is completely reasonable already.

Thanks for verifying.

PR #151, opened 2024-01-06
Replace OrderStatus with enum type

erinnmclaughlin:
This is related to #23, specifically  the "Use actual CLR enums for CardType and OrderStatus" task. 

This PR only addresses `OrderStatus` to keep changes minimal, but I'm happy to create a follow-up PR (or update this one) to address the `CardType` enum too.

A (perhaps bold) decision here was to drop the `orderstatus` table in favor of an `OrderStatus` column on the `orders` table. There's a migration associated with this, which includes some custom SQL to migrate the data. Please let me know if this is appropriate!

Also note that this PR will introduce conflicts with PR #115.


erinnmclaughlin:
@dotnet-policy-service agree

src/Ordering.API/Application/DomainEventHandlers/OrderCancelledDomainEventHandler.cs
```diff
@@ -22,12 +22,12 @@ public partial class OrderCancelledDomainEventHandler
...
     public async Task Handle(OrderCancelledDomainEvent domainEvent, CancellationToken cancellationToken)
     {
-        OrderingApiTrace.LogOrderStatusUpdated(_logger, domainEvent.Order.Id, nameof(OrderStatus.Cancelled), OrderStatus.Cancelled.Id);
+        OrderingApiTrace.LogOrderStatusUpdated(_logger, domainEvent.Order.Id, OrderStatus.Cancelled);
 
         var order = await _orderRepository.GetAsync(domainEvent.Order.Id);
         var buyer = await _buyerRepository.FindByIdAsync(order.GetBuyerId.Value);
 
-        var integrationEvent = new OrderStatusChangedToCancelledIntegrationEvent(order.Id, order.OrderStatus.Name, buyer.Name, buyer.IdentityGuid);
+        var integrationEvent = new OrderStatusChangedToCancelledIntegrationEvent(order.Id, order.OrderStatus.ToString(), buyer.Name, buyer.IdentityGuid);
```
captainsafia, line 30: Can we change the `OrderStatusChangedToCancelledIntegrationEvent` to take `OrderStatus` as an enum instead of a `string`?

captainsafia (commented):
Thanks for opening this PR! We'd certainly appreciate another one for `CardType` once this one is merged.

I ran into some issues while running our test flow locally:

```

      Fatal error establishing database connection
      Npgsql.PostgresException (0x80004005): 42703: column "OrderStatusId" does not exist
      
      POSITION: 82
         at Npgsql.Internal.NpgsqlConnector.ReadMessageLong(Boolean async, DataRowLoadingMode dataRowLoadingMode, Boolean readingNotifications, Boolean isReadingPrependedMessage)
         at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
         at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming, CancellationToken cancellationToken)
         at Npgsql.NpgsqlDataReader.NextResult(Boolean async, Boolean isConsuming, CancellationToken cancellationToken)
         at Npgsql.NpgsqlCommand.ExecuteReader(Boolean async, CommandBehavior behavior, CancellationToken cancellationToken)
         at Npgsql.NpgsqlCommand.ExecuteReader(Boolean async, CommandBehavior behavior, CancellationToken cancellationToken)
         at eShop.OrderProcessor.Services.GracePeriodManagerService.GetConfirmedGracePeriodOrders() in /Users/captainsafia/gh/eShop/src/OrderProcessor/Services/GracePeriodManagerService.cs:line 79
        Exception data:
          Severity: ERROR
          SqlState: 42703
          MessageText: column "OrderStatusId" does not exist
          Hint: Perhaps you meant to reference the column "orders.OrderStatus".
          Position: 82
          File: parse_relation.c
          Line: 3656
          Routine: errorMissingColumn
```

It looks like the [GracePeriodManagerService](https://github.com/dotnet/eShop/blob/main/src/OrderProcessor/Services/GracePeriodManagerService.cs) might need to be updated too.

captainsafia:
Also, in case it is helpful, the local test flow is to:

- Log in
- Add item to cart
- Update counter from 1 to 2
- Checkout
- Go to order status page
- Observe order status goes from submitted -> awaiting validation -> stock confirmed -> .. -> shipped

captainsafia:
Thanks for updating the GracePeriodManagerService! I noticed one other place in the codebase that we'll need to change in reaction to the new enum.

The orders UI currently sets the status as a CSS class:

https://github.com/dotnet/eShop/blob/63e82c9e75b1a50cd28499240fcbc2942bdc6939/src/WebApp/Components/Pages/User/Orders.razor#L40-L44

Which then gets stylized a certain way:

https://github.com/dotnet/eShop/blob/63e82c9e75b1a50cd28499240fcbc2942bdc6939/src/WebApp/Components/Pages/User/Orders.razor.css#L56-L64

It looks like the default `ToString` for the enum will render the string in PascalCase:

image

Which means that the style won't end up applying since CSS selectors are case sensitive.

I _think_ applying `ToString` on the line above should do the trick.

src/Ordering.API/Application/DomainEventHandlers/OrderCancelledDomainEventHandler.cs
erinnmclaughlin: Hey Safia, thanks for your feedback on this! I'm looking into making this change, but I noticed that we serialize & persist the event data, so `OrderStatus` would end up being serialized as an `int` instead of a `string` by default. 

How do you want to handle that? I could throw a `[[JsonConverter(typeof(JsonStringEnumConverter))]` over the enum? Ex:

```csharp
public record OrderStatusChangedToAwaitingValidationIntegrationEvent : IntegrationEvent
{
    public int OrderId { get; }
    [JsonConverter(typeof(JsonStringEnumConverter))]
    public OrderStatus OrderStatus { get; }
    public string BuyerName { get; }
    public string BuyerIdentityGuid { get; }
    public IEnumerable<OrderStockItem> OrderStockItems { get; }

    public OrderStatusChangedToAwaitingValidationIntegrationEvent(
        int orderId, OrderStatus orderStatus, string buyerName, string buyerIdentityGuid,
        IEnumerable<OrderStockItem> orderStockItems)
    {
        OrderId = orderId;
        OrderStockItems = orderStockItems;
        OrderStatus = orderStatus;
        BuyerName = buyerName;
        BuyerIdentityGuid = buyerIdentityGuid;
    }
}
```

captainsafia: @erinnmclaughlin I like that recommendation! I'm inclined to apply the attribute on the `public enum OrderStatus` definition itself so that it gets serialized consistently as a string throughout the codebase.

erinnmclaughlin: @captainsafia didn't even realize that was an option 😅 that is much nicer. Done!

captainsafia (approved):
Order flow works and CI is passing. Thanks for all your work on this, @erinnmclaughlin! 😄

PR #153, opened 2024-01-10
Aspire.AI.Azure.OpenAI integration

sebastienros:
Showing how the upcoming Aspire.Azure.AI.OpenAI component can be integrated.

### DIFF
--- a/src/Catalog.API/AIOptions.cs
+++ b/src/Catalog.API/AIOptions.cs
@@ -8,11 +8,10 @@ public class AIOptions
 
 public class OpenAIOptions
 {
-    /// <summary>OpenAI API key for accessing embedding LLM.</summary>
-    public string ApiKey { get; set; }
-
-    /// <summary>Optional endpoint for which OpenAI API to access.</summary>
-    public string Endpoint { get; set; }
+    /// <summary>
+    /// Gets or sets whether OpenAI services should be enabled.
+    /// </summary>
+    public bool Enable { get; set; } = true;
 
     /// <summary>The name of the embedding model to use.</summary>
     /// <remarks>When using Azure OpenAI, this should be the "Deployment name" of the embedding model.</remarks>
--- a/src/Catalog.API/Program.cs
+++ b/src/Catalog.API/Program.cs
@@ -5,6 +5,7 @@ builder.AddDefaultOpenApi();
 builder.AddApplicationServices();
 
 builder.Services.AddProblemDetails();
+builder.AddAzureOpenAI("catalogOpenAi");
 
 var app = builder.Build();
 
--- a/src/Catalog.API/Services/CatalogAI.cs
+++ b/src/Catalog.API/Services/CatalogAI.cs
@@ -1,38 +1,30 @@
-﻿using Azure;
-using Azure.AI.OpenAI;
+﻿using Azure.AI.OpenAI;
 using Pgvector;
 
 namespace eShop.Catalog.API.Services;
 
 public sealed class CatalogAI : ICatalogAI
 {
-    /// <summary>OpenAI API key for accessing embedding LLM.</summary>
-    private readonly string _aiKey;
-    /// <summary>Optional OpenAI API endpoint.</summary>
-    private readonly string _aiEndpoint;
-    /// <summary>The name of the embedding model to use.</summary>
     private readonly string _aiEmbeddingModel;
+    private readonly OpenAIClient _openAIClient;
 
     /// <summary>The web host environment.</summary>
     private readonly IWebHostEnvironment _environment;
     /// <summary>Logger for use in AI operations.</summary>
     private readonly ILogger _logger;
 
-    public CatalogAI(IOptions<AIOptions> options, IWebHostEnvironment environment, ILogger<CatalogAI> logger)
+    public CatalogAI(IOptions<AIOptions> options, OpenAIClient openAIClient, IWebHostEnvironment environment, ILogger<CatalogAI> logger)
     {
         var aiOptions = options.Value;
 
-        _aiKey = aiOptions.OpenAI.ApiKey;
-        _aiEndpoint = aiOptions.OpenAI.Endpoint;
         _aiEmbeddingModel = aiOptions.OpenAI.EmbeddingName ?? "text-embedding-ada-002";
-        IsEnabled = !string.IsNullOrWhiteSpace(_aiKey);
-
+        IsEnabled = aiOptions.OpenAI.Enable;
+        _openAIClient = openAIClient;
         _environment = environment;
         _logger = logger;
 
         if (_logger.IsEnabled(LogLevel.Information))
         {
-            _logger.LogInformation("API Key: {configured}", string.IsNullOrWhiteSpace(_aiKey) ? "Not configured" : "Configured");
             _logger.LogInformation("Embedding model: \"{model}\"", _aiEmbeddingModel);
         }
     }
@@ -54,18 +46,11 @@ public sealed class CatalogAI : ICatalogAI
         }
 
         EmbeddingsOptions options = new(_aiEmbeddingModel, [text]);
-        return new Vector((await GetAIClient().GetEmbeddingsAsync(options)).Value.Data[0].Embedding);
+        return new Vector((await _openAIClient.GetEmbeddingsAsync(options)).Value.Data[0].Embedding);
     }
 
     /// <summary>Gets an embedding vector for the specified catalog item.</summary>
     public ValueTask<Vector> GetEmbeddingAsync(CatalogItem item) => IsEnabled ?
         GetEmbeddingAsync($"{item.Name} {item.Description}") :
         ValueTask.FromResult<Vector>(null);
-
-    /// <summary>Gets the AI client used for creating embeddings.</summary>
-    private OpenAIClient GetAIClient() => !string.IsNullOrWhiteSpace(_aiKey) ?
-        !string.IsNullOrWhiteSpace(_aiEndpoint) ?
-            new OpenAIClient(new Uri(_aiEndpoint), new AzureKeyCredential(_aiKey)) :
-            new OpenAIClient(_aiKey) :
-        throw new InvalidOperationException("AI API key not configured");
 }
--- a/src/Catalog.API/appsettings.json
+++ b/src/Catalog.API/appsettings.json
@@ -26,7 +26,8 @@
   }
   //"AI": {
   //  "OpenAI": {
-  //    "APIKey": "",
+  //    "Enabled": true,
   //    "EmbeddingName": ""
   //  }
+  //}
 }
--- a/src/WebApp/Extensions/Extensions.cs
+++ b/src/WebApp/Extensions/Extensions.cs
@@ -1,11 +1,14 @@
-﻿using eShop.WebApp;
+﻿using Azure.AI.OpenAI;
+using eShop.WebApp;
 using eShop.WebAppComponents.Services;
 using Microsoft.AspNetCore.Authentication.Cookies;
 using Microsoft.AspNetCore.Authentication.OpenIdConnect;
 using Microsoft.AspNetCore.Components.Authorization;
 using Microsoft.AspNetCore.Components.Server;
 using Microsoft.IdentityModel.JsonWebTokens;
-using Microsoft.SemanticKernel;
+using Microsoft.SemanticKernel.ChatCompletion;
+using Microsoft.SemanticKernel.Connectors.OpenAI;
+using Microsoft.SemanticKernel.TextGeneration;
 
 public static class Extensions
 {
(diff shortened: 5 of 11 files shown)


src/WebApp/Extensions/Extensions.cs
```diff
@@ -91,17 +94,22 @@ public static void AddAuthenticationServices(this IHostApplicationBuilder builde
...
-            var kernelBuilder = builder.Services.AddKernel();
-            if (!string.IsNullOrWhiteSpace(openAIOptions.Endpoint))
-            {
-                kernelBuilder.AddAzureOpenAIChatCompletion(openAIOptions.ChatModel, openAIOptions.Endpoint, openAIOptions.ApiKey);
-            }
-            else
+            builder.Services.AddKernel();
+            builder.AddAzureOpenAI("chatOpenAi");
+
+            // Note: SemanticKernel should expose an extension method AddAzureOpenAIChatCompletion() that does the following.
```
sebastienros, line 104: Maybe there is already an extension for this.

src/Catalog.API/AIOptions.cs
```diff
@@ -8,11 +8,10 @@ public class AIOptions
...
 {
-    /// <summary>OpenAI API key for accessing embedding LLM.</summary>
-    public string ApiKey { get; set; }
-
-    /// <summary>Optional endpoint for which OpenAI API to access.</summary>
-    public string Endpoint { get; set; }
+    /// <summary>
+    /// Gets or sets whether OpenAI services should be enabled.
+    /// </summary>
+    public bool Enable { get; set; } = true;
```
stephentoub, line 14: Is this necessary? Can whether this is enabled be based on whether an OpenAIClient is available from DI, which would in turn be based on whether it was appropriately configured?

src/WebApp/Extensions/Extensions.cs
```diff
@@ -91,17 +94,22 @@ public static void AddAuthenticationServices(this IHostApplicationBuilder builde
...
+            // Currently these methods only allow to register an OpenAIClient they construct.
+            Func<IServiceProvider, object?, AzureOpenAIChatCompletionService> implementationFactory = delegate (IServiceProvider serviceProvider, object? _)
             {
-                kernelBuilder.AddOpenAIChatCompletion(openAIOptions.ChatModel, openAIOptions.ApiKey);
-            }
+                return new AzureOpenAIChatCompletionService(deploymentName, serviceProvider.GetRequiredService<OpenAIClient>(), null, serviceProvider.GetService<ILoggerFactory>());
+            };
+
+            builder.Services.AddKeyedSingleton(default(string), (Func<IServiceProvider, object, IChatCompletionService>)implementationFactory);
+            builder.Services.AddKeyedSingleton(default(string), (Func<IServiceProvider, object, ITextGenerationService>)implementationFactory);
```
stephentoub, line 112: The second line here isn't needed. The app only uses the IChatCompletionService.

src/WebApp/Extensions/Extensions.cs
```diff
@@ -91,17 +94,22 @@ public static void AddAuthenticationServices(this IHostApplicationBuilder builde
...
+            // Note: SemanticKernel should expose an extension method AddAzureOpenAIChatCompletion() that does the following.
+            // Currently these methods only allow to register an OpenAIClient they construct.
+            Func<IServiceProvider, object?, AzureOpenAIChatCompletionService> implementationFactory = delegate (IServiceProvider serviceProvider, object? _)
             {
-                kernelBuilder.AddOpenAIChatCompletion(openAIOptions.ChatModel, openAIOptions.ApiKey);
-            }
+                return new AzureOpenAIChatCompletionService(deploymentName, serviceProvider.GetRequiredService<OpenAIClient>(), null, serviceProvider.GetService<ILoggerFactory>());
+            };
+
+            builder.Services.AddKeyedSingleton(default(string), (Func<IServiceProvider, object, IChatCompletionService>)implementationFactory);
```
stephentoub, line 111: Why AddKeyedSingleton with a null key rather than AddSingleton?

src/WebApp/Extensions/Extensions.cs
stephentoub: There isn't currently. I wanted us to wait to add one until the aspire component was on the way such that it'd be more common to have one of these available.

src/Catalog.API/appsettings.json
```diff
@@ -17,6 +17,7 @@
   },
   "ConnectionStrings": {
     "EventBus": "amqp://localhost"
+    //"catalogOpenAi": "Endpoint=xxxx;Key=xxxx"
```
eerhardt, line 20: One nice thing about Aspire is that we can centralize this configuration in the eShop.AppHost's configuration. Thus a local dev only needs to set this info in 1 place (either `eShop.AppHost\appsetting.json` or eShop.AppHost's User Secrets) and it will flow to both Catalog.API and to WebApp.

So can we move this to `eShop.AppHost\appsetting.json`?

src/eShop.AppHost/Program.cs
```diff
@@ -14,6 +14,9 @@
 var orderDb = postgres.AddDatabase("OrderingDB");
 var webhooksDb = postgres.AddDatabase("WebHooksDB");
 
+var catalogOpenAi = builder.AddOpenAI("catalogOpenAi");
+var chatOpenAi = builder.AddOpenAI("chatOpenAi");
```
eerhardt, line 18: Why 2 separate ones? I assume we only want 1, and it flows to both projects.

src/WebApp/Extensions/Extensions.cs
stephentoub: https://github.com/microsoft/semantic-kernel/pull/4555

src/WebApp/Extensions/Extensions.cs
```diff
@@ -91,17 +94,22 @@ public static void AddAuthenticationServices(this IHostApplicationBuilder builde
...
+        if (!string.IsNullOrWhiteSpace(deploymentName))
         {
-            var kernelBuilder = builder.Services.AddKernel();
-            if (!string.IsNullOrWhiteSpace(openAIOptions.Endpoint))
-            {
-                kernelBuilder.AddAzureOpenAIChatCompletion(openAIOptions.ChatModel, openAIOptions.Endpoint, openAIOptions.ApiKey);
-            }
-            else
+            builder.Services.AddKernel();
+            builder.AddAzureOpenAI("chatOpenAi");
```
eerhardt, line 102: How does the app behave when the dev hasn't configured openai? Before if it wasn't configured the Kernel wasn't added to DI, and the app checks for this. See 

https://github.com/dotnet/eShop/blob/63e82c9e75b1a50cd28499240fcbc2942bdc6939/src/WebApp/Components/Chatbot/Chatbot.razor#L56-L57

src/eShop.AppHost/Program.cs
sebastienros: Just in case you'd want to use two instances. But we can make the assumption that a single service is used in this app.

src/WebApp/Extensions/Extensions.cs
sebastienros: That is the point of the `Enable` property since there is no Key property to check anymore. I just did it in one section.

How do we want it to work? `Enable` flag? Try to resolve the service?

src/Catalog.API/Program.cs
```diff
@@ -5,6 +5,7 @@
 builder.AddApplicationServices();
 
 builder.Services.AddProblemDetails();
+builder.AddAzureOpenAI("catalogOpenAi");
```
sebastienros, line 8: I don't like the name of this method. 
In the AppHost we have `AddAzureOpenAI` and `AddopenAI` (for non-azure and azure based open ai services).
In the apps we have `AddAzureOpenAI` and `AddKeyedAzureOpenAI` named after the NuGet package.

On the "client" app I'd rather have the method called `Add[Keyed]OpenAIClient` since is independent of Azure or non-Azure, and represents more what it's doing.

sebastienros: "Azure" is OK, we could see this coming from the Azure SDK.

What about the `AddOpenAI` vs `AddOpenAIClient` though? For Postgres we call it `AddNpgsqlDataSource` which makes it simpler to grasp, and why I'd prefer to use `AddOpenAIClient` (or `AddAzureOpenAIClient` since it's an Azure SDK concept?)

src/Catalog.API/AIOptions.cs
sebastienros: We used the ApiKey to detect if the service should be registered. This property is not set anymore since the Aspire component reads a connection string. We can use the existence of the connection string instead probably, and remove the `Enable` property of the other service too.

src/WebApp/Extensions/Extensions.cs
eerhardt: > but the AppHost fails if there is none defined

How does it fail? Will this get better when we change it to be "ConnectionProperties/Info" instead of "ConnectionString"?

We have an "optional" concept on WithReference. See https://github.com/dotnet/aspire/blob/360c6a9e25d43b71921584288c609032cedd9fcf/src/Aspire.Hosting/Extensions/ResourceBuilderExtensions.cs#L146

> How do we want it to work? Enable flag? Try to resolve the service?

My opinion is that we don't have an "Enable" flag, but instead detect if the `Key` is configured.

src/Catalog.API/Program.cs
eerhardt: I'm OK with adding `Client` to the name.

src/WebApp/Extensions/Extensions.cs
stephentoub: > And as you can see this is what it's doing (AddKeyedSingleton with a null key). I didn't want to change anything.

It's not. It's sound AddKeyedSingleton with the supplied key, which may or may not be null.

src/Catalog.API/Program.cs
KrzysztofCwalina: FYI: we are having lots of discussions on the Azure SDK about the best way to support both OpenAI (OAI) and Azure OpenAI (AOAI) with just one client. We are thinking that the way it's done today (e.g. if no endpoint passed, OAI is called; and the 3rd parameter meaning either model or deployment) is not ideal and so it might change before the SDK GAs (releases as stable). Because of that, I would suggest that you don't rely too much on how the ctors look today.

stephentoub: FYI, depending on how that's changed, changing that could negatively impact Semantic Kernel, which has OpenAIClient in its public surface area.
cc: @markwallace-microsoft, @matthewbolanos

eerhardt:
What's the status of this PR? Are we waiting for Aspire preview3 to ship before completing it?

adityamandaleeka:
@sebastienros, FYI we're using preview 3 in this repo now (and you have a merge conflict for that).

src/eShop.AppHost/Program.cs
```diff
@@ -18,6 +18,8 @@
 var orderDb = postgres.AddDatabase("orderingdb");
 var webhooksDb = postgres.AddDatabase("webhooksdb");
 
+var openAi = builder.AddAzureOpenAI("OpenAi");
```
eerhardt, line 21:
```suggestion
var openAi = builder.AddAzureOpenAI("openai");
```

Everything else seems to be in all lower case.

src/Catalog.API/Services/CatalogAI.cs
```diff
@@ -1,38 +1,29 @@
...
     private readonly string _aiEmbeddingModel;
+    private readonly OpenAIClient _openAIClient;
 
     /// <summary>The web host environment.</summary>
     private readonly IWebHostEnvironment _environment;
     /// <summary>Logger for use in AI operations.</summary>
     private readonly ILogger _logger;
 
-    public CatalogAI(IOptions<AIOptions> options, IWebHostEnvironment environment, ILogger<CatalogAI> logger)
+    public CatalogAI(IOptions<AIOptions> options, IEnumerable<OpenAIClient> openAIClient, IWebHostEnvironment environment, ILogger<CatalogAI> logger)
```
eerhardt, line 16:
```suggestion
    public CatalogAI(IOptions<AIOptions> options, OpenAIClient openAIClient, IWebHostEnvironment environment, ILogger<CatalogAI> logger)
```

Why is this IEnumerable? I don't think it should be.

eerhardt (commented):
This looks good. I just have 1 main question.

src/Catalog.API/Services/CatalogAI.cs
sebastienros: To check if the service is actually registered.

eerhardt: Then make it:

```suggestion
    public CatalogAI(IOptions<AIOptions> options, IWebHostEnvironment environment, ILogger<CatalogAI> logger, OpenAIClient? openAIClient = null)
```

eerhardt (approved):

PR #197, opened 2024-02-17
added azure cicd via gha

bradygaster:


jamesmontemagno (approved):

danmoseley (approved):
Fine if it works

adityamandaleeka:
@bradygaster Can we have this instead spit the YAML from the CI job and then have it run provision and deploy using that?

jamesmontemagno:
@bradygaster updates if you are able to split this per ^

isaaclevintest:
@adityamandaleeka are we going to have this PR in the base repo? If yes, are there going to be a doc to define the pre-reqs to get this to work?

isaacrlevin:
@adityamandaleeka are we going to have this PR in the base repo? If yes, are there going to be a doc to define the pre-reqs to get this to work?

bradygaster:
@jamesmontemagno how should i split it? provision and deploy as separate files/jobs?

adityamandaleeka:
@bradygaster I think there was a typo above. The change we want to make is to _spit_ (emit) the YAML in the CI job itself and then use it in the subsequent parts. The script/flow you have already is good 😄 

@isaacrlevin Not sure what you mean by base repo (I assume you mean this repo?). This is just a CI job but we do have instructions for azd deployment in the README: https://github.com/dotnet/eShop?tab=readme-ov-file#use-azure-developer-cli

isaacrlevin:
I was just curious if we plan to actually have a version of eShop in Azure. If not, do we need to have the Action in the repo at all.

adityamandaleeka:
I think it's a good way for us to ensure we don't regress the ability to  deploy a 'vanilla' Aspire app to Azure.

jamesmontemagno:
Should we actually deploy it? if we do should we `azd down` at the end? what sub would we put this in?

bradygaster:
Not sure we should `azd down` at the end. Seems like if someone's setting it up they'd know they'll need to tear it down later.

isaacrlevin:
Probably should just put this in a sample or doc on "one of the ways" to deploy Aspire to Azure. That way you can outline steps to add secrets to repo and add a "Delete Resources` section that calls out `azd down`

PR #218, opened 2024-02-28
Replaced Map forwarder with Yarp reverse Proxy for mobile BFF

SrushtiPasari94:
Replaced Map forwarder with Yarp reverse Proxy for mobile BFF

**Status : under testing**


samsp-msft:
Is there a reason you have so many paths in the config that all point to the same cluster - is this so that the proxy will only forward those specific requests - you could handle many of those with a wildcard.

Eilon:
> Is there a reason you have so many paths in the config that all point to the same cluster - is this so that the proxy will only forward those specific requests - you could handle many of those with a wildcard.

I think we specifically want to avoid a wildcard because then it could accidentally map "too much". Instead, we want to map only exactly what we know we need so that we don't expose too much surface area of a back-end API.

jamesmontemagno:
Thanks @Eilon that helps. so if that is the case I think this is okay if testing goes well.

adityamandaleeka (approved):

adityamandaleeka:
@Eilon do you plan to review this?

jamesmontemagno (approved):
@SrushtiPasari94 testing going good?

src/Mobile.Bff.Shopping/Program.cs
```diff
@@ -3,36 +3,15 @@
...
-    CatalogApiPrefix + "items/type/all/brand/{brandId?}",
-    CatalogApiPrefix + "catalogTypes",
-    CatalogApiPrefix + "catalogBrands",
-
-    CatalogApiPrefix + "items/{id}/pic",
-};
-
-foreach (var forwardedUrl in forwardedCatalogApis)
-{
-    var mapFromPattern = "/catalog-api" + forwardedUrl;
```
Eilon, line 32: The only thing I don't see in the proxy config now is the `/catalog-api` prefix that I think is used for the 'local name' of the API. That is, the Mobile App (e.g. .NET MAUI Blazor Hybrid app) talk to the Mobile BFF at:

```
http://mobile-bff/catalog-api/<some catalog path>
                  ^^^^^^^^^^^
```

But I don't see where the new proxy config has that? Or is it not needed?

Ultimately just running the .NET MAUI Blazor Hybrid app (the `HybridApp` folder in the repo) will either work or not work, but my inclination is that it won't work because this path doesn't exist. Unless the proxy config just figures that out somehow?

jamesmontemagno: hmmmm let me take a look, I thought this would do it - "Address": "http://catalog-api/api/v1/catalog""

jamesmontemagno:
@Eilon  this should address this :)

Eilon:
Thanks, that looks more like what I would have expected, but I'm not sure how exactly YARP maps things, so I wasn't sure.

Ultimately, does the HybridApp project run and is able to make API calls to load the catalog? If so, good. If not, back to work! 😁

jamesmontemagno:
@Eilon yup! working good here :) The port is different for some reason, but beyond that good to go.

image

PR #233, opened 2024-03-10
refactor: avoid closure creation

Ted-Zhang:
refactor: avoid closure creation

The ```ExecutionStrategyExtensions.ExecuteAsync``` has an overload method to support passing the state, thereby updating the usage to avoid creating the closure to capture the reference.

Since this code example is part of the official documentation, so I think providing a refined code example would be beneficial.
[Official document link](https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/implement-resilient-entity-framework-core-sql-connections)

### DIFF
--- a/src/IntegrationEventLogEF/Utilities/ResilientTransaction.cs
+++ b/src/IntegrationEventLogEF/Utilities/ResilientTransaction.cs
@@ -13,9 +13,9 @@ public class ResilientTransaction
         //Use of an EF Core resiliency strategy when using multiple DbContexts within an explicit BeginTransaction():
         //See: https://docs.microsoft.com/en-us/ef/core/miscellaneous/connection-resiliency
         var strategy = _context.Database.CreateExecutionStrategy();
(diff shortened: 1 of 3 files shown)


Ted-Zhang:
@dotnet-policy-service agree

BrennanConroy:
The closures capture other variables so I don't think you're really saving anything here. You can check to see if the closure is avoided by making the delegate static.
```csharp
int i;
F(static func()
{
    Console.WriteLine(i); // compiler error, i is captured but delegate is static
});

void F(Action a);
```

Ted-Zhang:
The ```IExecutionStrategy``` from EF Core has an overloaded method that allows passing the status, this is leveraging that to avoid the closure creation to capture those variables. 
https://github.com/dotnet/efcore/blob/fa093d46c94d490317ee87ac257ed5f37a3f5565/src/EFCore/Storage/ExecutionStrategyExtensions.cs#L228C1-L242C1

PR #267, opened 2024-03-27
Only enable Swagger UI under development mode

captainsafia:
Only enable Swagger UI for local development as a best practice.


danmoseley (approved):
I assume we'll have a breaking change doc synced with this?

captainsafia:
> I assume we'll have a breaking change doc synced with this?

Assuming I understand what you mean, I don't think we document breaking changes on eShop? Our templates already guard the `UseSwaggerUi` calls so this isn't a breaking change for anything we ship out of ASP.NET Core.

danmoseley:
Yeah, I meant for equivalent change in templates but didn't check. Sounds good

PR #269, opened 2024-03-29
Fix WebApps OpenAI usage

eerhardt:
The Kernel was never registered in DI, so it wasn't possible to use the AI services.

### DIFF
--- a/src/WebApp/Extensions/Extensions.cs
+++ b/src/WebApp/Extensions/Extensions.cs
@@ -103,6 +103,7 @@ public static class Extensions
 
         if (!string.IsNullOrWhiteSpace(builder.Configuration.GetConnectionString("openai")) && !string.IsNullOrWhiteSpace(deploymentName))
         {
+            builder.Services.AddKernel();
             builder.AddAzureOpenAI("openai");
             builder.Services.AddAzureOpenAIChatCompletion(deploymentName);
         }


src/WebApp/Extensions/Extensions.cs
```diff
@@ -103,6 +103,7 @@ private static void AddAIServices(this IHostApplicationBuilder builder)
 
         if (!string.IsNullOrWhiteSpace(builder.Configuration.GetConnectionString("openai")) && !string.IsNullOrWhiteSpace(deploymentName))
         {
+            builder.Services.AddKernel();
```
sebastienros, line 106: Looking at my PR (which removed it, probably by mistake)_it was then registering the services on the returned builder `var kernelBuilder = builder.Services.AddKernel();`. I check the source and it's the same as `builder.Service` so this is fine.

sebastienros (approved):

eerhardt:
@adityamandaleeka - the Playwright tests are failing. I'm pretty sure I didn't break them. Is this a known problem?

adityamandaleeka:
Tagging @SrushtiPasari94 for the failing playwright tests.

SrushtiPasari94:
@adityamandaleeka Seems test are not failing, and PR is merged now.

Issue #21, opened 2023-11-01
Use API versioning for HTTP-based services


adityamandaleeka:
From @captainsafia

See https://github.com/dotnet/aspnet-api-versioning

commonsensesoftware:
Hey @captainsafia! Great to hear from you. Congrats on the release. Hope all is well.

I'd love to help out with this. If there are any specific requirement, features, or behaviors that need to be showcased, please call them out. I ran into a few snags building the `net8.0` TFM, but I think I've got them resolved and I expect to have a .NET 8 release by EOW. Do you have a specific deadline?

This would be great way to highlight one of the most under-utilized features, which is the API version-aware `HttpClient`. 😉 

https://github.com/dotnet/aspnet-api-versioning/wiki/Versioned-Clients

The [Asp.Versioning.Http.Client](https://www.nuget.org/packages/Asp.Versioning.Http.Client) package and features have been around for about two years, but they don't get used a whole lot - yet. It wires things up with DI extensions and `IHttpClientFactory`, resulting in the rest of the code not knowing or understanding which API version it's talking to. The `IApiVersionWriter` is the reciprocal of `IApiVersionReader` and will automatically add the API version to the query string, header, media type, or even the request URL. 😄

For example:

```c#
var response = await client.GetAsync("api/order/42"); // GET api/order/42?api-version=1.0
```

There some additional interesting features if the server reports more information, such as:

- Is an API version newer than the one you are talking to available?
- Is the API version you are talking to deprecated?
- If you're talking to a deprecated API version, when will to be permanently sunset?

This all abstracted behind `IApiNotification` with the default implementation plumbing through `ILogger`, but this would be a good opportunity to exercise those capabilities and see if any additional improvements are warranted.

Thanks for reaching out. Let me know.

captainsafia:
> but I think I've got them resolved and I expect to have a .NET 8 release by EOW. Do you have a specific deadline?

No specific deadline! Integrating this after the .NET 8 release of the package is out should be sufficient.

> This would be great way to highlight one of the most under-utilized features, which is the API version-aware HttpClient. 

I think it would be interesting to try showcasing the versioned HTTP client here, if we can get it wired up with the authentication and service discovery configuration on our current setup.

> There some additional interesting features if the server reports more information, such as:

Interesting features. At the moment, we only have a single version of the API so there's not much to maintain with regard to different versions.

I think we can start with a basic versioning implementation on the APIs + versioned HTTP client for this.

PR #272, opened 2024-03-30
Integrate ASP.NET API Versioning

commonsensesoftware:
## Overview

At the request of #21, this PR adds support for API Versioning to the reference application.

## Additions

- APIs support versions
- All API versions are explicit
- The existing, default API version `1.0` is retained
  - It might be more practical to switch from the numeric to the date format
- API Explorer extensions have been added to discover and include versioned API information
- Swashbuckle customization has been included to leverage all of the versioned API information
- `HttpClient` extensions are included to integrate the target API version in outgoing requests

## Changes

The application fundamentally changes from versioning by URL segment to versioning by query string parameter. Without ASP.NET Versioning in place, I understand why that approach was employed. Versioning by URL segment, despite being very popular, is not RESTful; it violates the _Uniform Interface_ constraint. A resource can have multiple identifiers, but that is not what `v1/catalog/item/42` versus `v2/catalog/item/42` implies. The expectation is a different representation versus a different resource. Representations are indicated by a media type in `Accept` or `Content-Type`. A query string parameter is a pragmatic balance. It doesn't change the resource identifier (e.g. the _path_), but is still _easy_ to indicate API versions without having to set header values.

This is the just the tip of the iceberg of problems that can arise for versioning by URL segment. HATEOAS becomes nearly impossible or coupled as APIs evolve independently over time. In a self-contained application, where both sides are owned, that might not be an issue. If this application were to hypothetically open up to external callers (ex: for consignment), there could be challenges when linking resources. As an example, if `order/abc` had an item with a link to `catalog/item/42` and there are two versions, how would the server know which version to pick? It might assume the same version as the service itself or the incoming request, but those are both flawed. There is no requirement or guarantee a client uses the same version across all APIs. This _can_ be facilitated by enforcing _symmetric versioning_ across all APIs, but that requires adding new API versions when no changes have occurred. Ultimately, the client should **always** explicitly specify the API version they want (much in the same way that `Accept` indicates what format they want it in).

As a reference application, this example should provide a template that leads developers to the _Pit of Success_. Other arguments aside, switching to the query string approach has the following additional benefits:

- Alignment to how [Azure APIs are versioned](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md)
- Alignment to the recommended [API Versioning examples](https://github.com/dotnet/aspnet-api-versioning/tree/main/examples/AspNetCore/WebApi/MinimalOpenApiExample)
- URLs remain stable across versions, which leads to:
   - Fewer changes
   - Less source control churn
   - HATEOAS, if every needed or desired

## Future Enhancements

After these changes are integrated, a few other enhancements are possible and potentially interesting:

- Adding the requested API version in instrumentation telemetry
- Discovery and reverse proxy mapping configuration from API version metadata (but no idea how - yet)

-----

cc: @captainsafia


src/Catalog.API/Catalog.API.http
```diff
@@ -1,10 +1,10 @@
 ﻿@Catalog.API_HostAddress = http://localhost:5222
 
 
-GET {{Catalog.API_HostAddress}}/api/v1/catalog/items
+GET {{Catalog.API_HostAddress}}/api/catalog/items?api-version=1.0
```
captainsafia, line 4: Should we use a variable to capture the API version here?

```suggestion
GET {{Catalog.API_HostAddress}}/api/catalog/items{{targetVersion}}
```

tests/Catalog.FunctionalTests/CatalogApiTests.cs
```diff
@@ -123,7 +127,7 @@ public async Task GetCatalogItemWithId()
     public async Task GetCatalogItemWithExactName()
     {
         // Act
-        var response = await _httpClient.GetAsync("api/v1/catalog/items/by/Wanderer%20Black%20Hiking%20Boots?PageSize=5&PageIndex=0");
+        var response = await _httpClient.GetAsync("api/catalog/items/by/Wanderer%20Black%20Hiking%20Boots?PageSize=5&PageIndex=0");
```
captainsafia, line 130: Love this delta!

src/Webhooks.API/Program.cs
```diff
@@ -1,18 +1,18 @@
...
 
 builder.AddServiceDefaults();
-builder.AddDefaultOpenApi();
 builder.AddApplicationServices();
 
-var app = builder.Build();
+var apiVersioning = builder.Services.AddApiVersioning();
 
-app.UseDefaultOpenApi();
+builder.AddDefaultOpenApi(apiVersioning);
```
captainsafia, line 8: Is there a reason we have to pass the `IApiVersioningBuilder` via the arguments in this way? It seems to me that we could possible call `AddApiVersioning` inside the `AddDefaultOpenApi` call.

src/Ordering.API/Program.cs
```diff
@@ -1,19 +1,19 @@
...
 
 app.MapDefaultEndpoints();
 
-app.MapGroup("/api/v1/orders")
-   .MapOrdersApi()
-   .RequireAuthorization();
+app.NewVersionedApi("Orders")
+   .MapOrdersApiV1();
 
+app.UseDefaultOpenApi();
```
captainsafia, line 18: Nit: Can we keep the `UseDefaultOpenApi` call before the `MapX` calls for consistency?

src/Ordering.API/Program.cs
```diff
@@ -1,19 +1,19 @@
...
+var apiVersioning = builder.Services.AddApiVersioning();
 
-app.UseDefaultOpenApi();
+builder.AddDefaultOpenApi(apiVersioning);
+
+var app = builder.Build();
 
 app.MapDefaultEndpoints();
 
-app.MapGroup("/api/v1/orders")
```
captainsafia, line 15: Should we keep the `RequireAuthorization` call in `Program.cs`? It strikes me that it's a useful piece of information regardless of the version of the API used.

src/Catalog.API/Catalog.API.http
commonsensesoftware: We _can_. I followed the existing theme here. There wasn't a variable for `v{{1}}`, but there certainly can be. It _might_ convenient in this scenario to run each API. In a scenario the services evolve independently over time, it's more likely API versions will be heterogenous and, therefore, a single API version parameter is less useful.

It's a simple change that I'm happy to make if you'd like to see it added.

src/Webhooks.API/Program.cs
commonsensesoftware: The reason that `IApiVersioningBuilder` is passed here has to do with how API Versioning is configured. As an example, `IMvcCoreBuilder.AddApiExplorer` is different than `IApiVersioningBuilder.AddApiExplorer`, but they intentionally have the same name as they express the same intention. Under the hood, the later calls the former.

It would be possible to **not** pass the `IApiVersioningBuilder` and have a setup inside `AddDefaultOpenApi` that looks like:

```c#
builder.Services.AddApiVersioning().AddApiExplorer();
```

There's a few reasons why I chose not to do that:

1. Ordering can matter. If `AddApiVersioning()` hasn't already been called and configured, then services _might_ not be configured as expected. It's an edge case not likely to happen in a reference application, but someone copying and modifying it _could_ run into that problem in a very non-obvious way.
2. The name `AddDefaultOpenApi` implies that is configuring OpenAPI. If it also adds API Versioning, that is neither self-describing nor intrinsically obvious.
3. `AddDefaultOpenApi` is a shared extension method. It's possible that different APIs might configure API Versioning different ways; for example, different styles, different defaults, etc. It's better to the keep the API Versioning configuration separate so that it is obvious what and when it is configured. The API Versioning extensions for OpenAPI tend to be the same or overlap for all APIs so keeping it inside `AddDefaultOpenApi` makes sense. It may not even be invoked. By passing `IApiVersioningBuilder`, the ordering problem is solved because you had to have called `AddApiVersioning()` at least once.

I'll let you cogitate on that and decide if you'd like to change it.

src/Ordering.API/Program.cs
commonsensesoftware: This one I was less sure on. If I'm not mistaken, adding `RequireAuthorization` will recursively apply to all endpoints under the associated group; regardless, of the depth of the group. That seems straight forward and makes sense. I was wondering whether all versions of an API would use the same authorization. Probably? 🤷🏽‍♂️

If you don't see any other issues with having `RequireAuthorization` in `Program.cs`, I can revert it. A future scenario I considered that would fail in a not-so-obvious way was:

```c#
app.NewVersionedApi("Orders")
   .MapOrdersApiV1()        // ← no authz
   .MapOrdersApiV2()
   .RequireAuthorization(); // ← applies to V2 only
```

If the order is reversed, I _believe_ it will work:

```c#
app.NewVersionedApi("Orders") // ← empty, top-level group
   .RequireAuthorization()    // ← applies from top-level down, recursively
   .MapOrdersApiV1()          // ← v1 group under api with authz
   .MapOrdersApiV2();         // ← v2 group under api with authz
```

src/Ordering.API/Program.cs
commonsensesoftware: Unfortunately - no. This _limitation_ was one of the most infuriating part of _Minimal APIs_ (for me) when they were introduced. I understand why it works the way it does, but when all of the pieces are put together, things only work if configured in a very specific order. The issues are a perfect storm of all the libraries involved.

### Issue 1

ASP.NET Core backs `WebApplication` with an `EndpointDataSource` that it puts into the DI container. This is great for deferred evaluation since the backig services are designed to handle changes to the underlying endpoint collections.

https://github.com/dotnet/aspnetcore/blob/dc1acba9cd1374a8a8560bee655682e1a72de3eb/src/Http/Routing/src/DependencyInjection/RoutingServiceCollectionExtensions.cs#L75

### Issue 2

The API Explorer extensions provided API Versioning define the `IApiVersionDescriptionProvider` service, which is responsible for collating all of the API versions in the application. If all of the `Endpoint`s are deprecated, it will also handle reporting that a given API version is deprecated. This service is used to enumerate all API versions in an application. A common usage scenario is dynamically creating an OpenAPI document for each API version.

The `IApiVersionDescriptionProvider` service is backed by a collection of `IApiVersionMetadataCollationProvider` services that know how to collate API versions across _Minimal APIs_ and good 'ol controllers. These are all in the DI container and backed by the registered `EndpointDataSource` and `IActionDescriptorCollectionProvider`, respectively. At this point, there are no `Endpoint`s registered with `EndpointDataSource`; therefore, the _Minimal API_ collator doesn't discover anything. These services are aware that the underlying sources can change and will adapt, but the changes have to happen first.

https://github.com/dotnet/aspnet-api-versioning/blob/22a775b8eae36147587faaa062d7555260c94736/src/AspNetCore/WebApi/src/Asp.Versioning.Mvc.ApiExplorer/IEndpointRouteBuilderExtensions.cs#L27

### Issue 3

When you finally call `UseSwaggerUI()`, Swashbuckle uses eager evaluation. It's building the middleware for the pipeline and the DI container has already been built, so it isn't that unreasonable.

https://github.com/domaindrivendev/Swashbuckle.AspNetCore/blob/8f363f7359cb1cb8fa5de5195ec6d97aefaa16b3/src/Swashbuckle.AspNetCore.SwaggerUI/SwaggerUIBuilderExtensions.cs#L34

As a result, when the setup for `SwaggerUIOptions` is evaluated in `UseSwaggerUI`:

1. there are no _Minimal APIs_ registered
2. API Versioning doesn't see or collate any API versions from _Minimal APIs_ - yet
3. one or more OpenAPI documents are not generated (which is wrong and unexpected)

-----

You might be wondering why this hasn't been a more prevalent issue. If you're **not** using API Versioning, then it's not a problem because there's only a single OpenAPI document or you've hard-coded the mapping in the configuration setup. Once you want to use the API Versioning API Explorer extensions to drive the configuration automatically from the application, things start to fall down if they are not done in the correct order.

I've had more than one customer file an issue about this behavior. I always direct them to the API Versioning example projects. The differences in the example setups versus the vanilla out-of-the-box templates are only meant to vary by what is absolutely necessary to get things running, not my personal dogma about how I think the template should be structured.

From my experience, combining all of these concepts together requires `UseDefaultOpenApi ` be the last thing or nearly the last thing in the configuration. It definitely needs to be after `MapXXX` calls because that is the only way for API Versioning to discover them (automatically).

We wont' solve the problem here, but I fully support making this experience better. The order should not really matter. This behavior is something you probably want to account for in the forthcoming OpenAPI extensions that are planned for .NET 9.

src/Ordering.API/Program.cs
commonsensesoftware: Partially reverting this back. I think the way we _recommend_ this should be done is like this:

```c#
var orders = app.NewVersionedApi("Orders");

orders.MapOrdersApiV1()
      .RequireAuthorization();

// if 2.0 was added, this is what it would look like
orders.MapOrdersApiV2()
      .RequireAuthorization();
```

It's not a hard requirement, but that makes things clear and separated.

commonsensesoftware: Partially reverting this back. I think we want to _recommend_ that it be configured like so:

```c#
var orders = app.NewVersionedApi("Orders");

orders.MapOrdersApiV1()
      .RequireAuthorization();

// if 2.0 is added, it would look like this
orders.MapOrdersApiV2()
      .RequireAuthorization();
```

This is not a hard requirement, but it keeps things separate and clear.

src/Catalog.API/Catalog.API.http
commonsensesoftware: I've added a parameter as requested

src/Webhooks.API/Program.cs
commonsensesoftware: I've restructured this just a bit to make it read more clearly and retain the original functionality.

### Adding OpenAPI with API Versioning

```c#
var withApiVersioning = builder.Services.AddApiVersioning();
builder.AddDefaultOpenApi(withApiVersioning);
```

### Adding OpenAPI without API Versioning (Original)

```c#
builder.AddDefaultOpenApi();
```

src/eShop.ServiceDefaults/ConfigureSwaggerOptions.cs
```diff
@@ -0,0 +1,178 @@
...
+
+    private static string BuildDescription(ApiVersionDescription api, string description)
+    {
+        var text = new StringBuilder(description);
+
+        if (api.IsDeprecated)
+        {
+            if (text.Length > 0)
+            {
+                if (text[^1] == '.')
```
qin-guan, line 64: Hey! Spotted this PR while searching on how to implement versioning. I believe this should be `!=`?
```suggestion
                if (text[^1] != '.')
```

src/eShop.ServiceDefaults/ConfigureSwaggerOptions.cs
```diff
@@ -0,0 +1,178 @@
...
+                foreach (var link in policy.Links.Where(l => l.Type == "text/html"))
+                {
+                    text.AppendLine();
+
+                    if (link.Title.HasValue)
+                    {
+                        text.Append(link.Title.Value).Append(": ");
+                    }
+
+                    text.Append(link.LinkTarget.OriginalString);
```
qin-guan, line 102: Small quality of life improvement to make the link clickable!
```suggestion
                    text.Append($"<{link.LinkTarget.OriginalString}>");
```

src/eShop.ServiceDefaults/ConfigureSwaggerOptions.cs
commonsensesoftware: Good catch. Fixed.

src/eShop.ServiceDefaults/ConfigureSwaggerOptions.cs
commonsensesoftware: Nice suggestion. I don't think adding `<` and `>` alone will do the trick; however, I've refactored the implementation to render an unordered list of links, if any. If any links for HTML pages are defined, it will render like so:

```html
<h4>Links</h4>
<ul>
 <li><a href="policy.html">Versioning Policy</a></li>
</ul>
```

The [API Versioning examples](https://github.com/dotnet/aspnet-api-versioning/tree/main/examples) for OpenAPI have been updated to reflect this change as well.

src/Webhooks.API/Program.cs
captainsafia: @commonsensesoftware Thanks for sharing this! I have a better understanding of the problem here after taking a look at what it would take to integrate Asp.Versioning with the new built-in support in Microsoft.AspNetCore.OpenApi. We can discuss this further next week but for now I appreciate why this is setup this way.

captainsafia (approved):
Approving for now.

Some of this code will need to change once we update eShop to use the new APIs in `Microsoft.AspNetCore.OpenApi` for OpenAPI document generation.

We can track the changes that will need to be made in a separate issue as part of the work.

Thanks for opening this PR and for your patience here, @commonsensesoftware!

captainsafia:
@commonsensesoftware It looks like CI wasn't running on this before due to the merge conflicts. It looks like `ApiVersioningVersion` isn't explicit defined anywhere. I assume we'll want to target `8.1.0` for all packages?

commonsensesoftware:
@captainsafia yes, it should be `8.1.0`. there was a merge conflict with the original submission and I rebased in the upstream. I goofed something up during the fix. I tested it ran as expected so I'm not sure how this happened. Something was cached I guess. I can add that in quickly if you like.

commonsensesoftware:
Added. Hopefully the CI build runs and is successful this time.

PR #275, opened 2024-04-01
Fix product image urls in chat responses

stephentoub:
Over time it seems changes have been made such that product image links are no longer stored with the products, and as such chat responses no longer include pictures. This fixes it so that product images now again show up in chat responses.


eerhardt (approved):
The changes look reasonable to me.

Issue #280, opened 2024-04-04
Can't deploy to Azure with AI support


aaronpowell:
_Cross-linking from https://github.com/dotnet/aspire/issues/3409 for visibility_.

Due to the above mentioned bug, it's not possible to deploy to Azure if you have AI support enabled. I've worked around it by changing the AI setup code like this:

```csharp
bool useOpenAI = true;
if (useOpenAI)
{
    const string openAIName = "openai";
    const string textEmbeddingName = "text-embedding-ada-002";
    const string chatModelName = "gpt-35-turbo-16k";

    // to use an existing OpenAI resource, add the following to the AppHost user secrets:
    // "ConnectionStrings": {
    //   "openai": "Key=<API Key>" (to use https://api.openai.com/)
    //     -or-
    //   "openai": "Endpoint=https://<name>.openai.azure.com/" (to use Azure OpenAI)
    // }
    if (builder.Configuration.GetConnectionString(openAIName) is not null)
    {
        var openAI = builder.AddConnectionString(openAIName);

        catalogApi
            .WithReference(openAI)
            .WithEnvironment("AI__OPENAI__EMBEDDINGNAME", textEmbeddingName);

        webApp
            .WithReference(openAI)
            .WithEnvironment("AI__OPENAI__CHATMODEL", chatModelName); ;
    }
    else
    {
        // to use Azure provisioning, add the following to the AppHost user secrets:
        // "Azure": {
        //   "SubscriptionId": "<your subscription ID>"
        //   "Location": "<location>"
        // }
        var chatAIResource = builder.AddAzureOpenAI(openAIName)
            .AddDeployment(new AzureOpenAIDeployment(chatModelName, "gpt-35-turbo", "0613"));

        var embeddingAIResource = builder.AddAzureOpenAI($"{openAIName}-embedding")
            .AddDeployment(new AzureOpenAIDeployment(textEmbeddingName, "text-embedding-ada-002", "2"));

        catalogApi
            .WithReference(chatAIResource)
            .WithReference(embeddingAIResource)
            .WithEnvironment("AI__OPENAI__EMBEDDINGNAME", textEmbeddingName);

        webApp
            .WithReference(chatAIResource)
            .WithReference(embeddingAIResource)
            .WithEnvironment("AI__OPENAI__CHATMODEL", chatModelName);
    }
}
```

It's only a quick solution, as I don't think both services require chat + embeddings (and you need to update the services to be aware of the additional resource too).

[...]

PR #297, opened 2024-04-08
Using Semantic Kernel Memory

### LINKED ISSUE #282: Leverage Semantic Kernel memory for embeddings
Semantic Kernel has [memories](https://learn.microsoft.com/en-us/semantic-kernel/memories/) which allows you to store data in a vector database (such as Postgres) and then retrieve it using semantic functions that SK understands.

I think it would be better to show off the workflow of using memory rather than the custom approach to managing embeddings and vector queries that is in eShop.

_Note: Memory in SK is currently marked as Experimental, so it does pose a risk on adoption._

aaronpowell:
This moves from the custom solution for vector searching on Postgres, instead using the SK memory feature to do it. Added some new SK dependencies for the memory (and PG memory store), then removed the Embedding column from the current data model, as there is a new table with all that in it.

Refactored the seed logic to load the memory store using the previously generated embeddings.

Changed the CatalogAPI route to use the ISemanticTextMemory search feature to search memory, rather than the custom SQL query. This does mean we don't get distance surfaced, also, pagination is currently lost and SK memory doesn't support that (we could roll that ourselves if we want).

Included a fix so that AOAI can be deployed (issue #280), and pgadmin for easier debugging of the data in the database.

Fixes #282

### DIFF
--- a/src/Catalog.API/Apis/CatalogApi.cs
+++ b/src/Catalog.API/Apis/CatalogApi.cs
@@ -1,5 +1,4 @@
 ﻿using Microsoft.AspNetCore.Http.HttpResults;
-using Pgvector.EntityFrameworkCore;
 
 namespace eShop.Catalog.API;
 
@@ -15,7 +14,7 @@ public static class CatalogApi
         app.MapGet("/items/{catalogItemId:int}/pic", GetItemPictureById);
 
         // Routes for resolving catalog items using AI.
-        app.MapGet("/items/withsemanticrelevance/{text:minlength(1)}",  GetItemsBySemanticRelevance);
+        app.MapGet("/items/withsemanticrelevance/{text:minlength(1)}", GetItemsBySemanticRelevance);
 
         // Routes for resolving catalog items by type and brand.
         app.MapGet("/items/type/{typeId}/brand/{brandId?}", GetItemsByBrandAndTypeId);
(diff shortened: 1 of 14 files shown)


src/Catalog.API/Apis/CatalogApi.cs
```diff
@@ -126,38 +125,21 @@ public static IEndpointRouteBuilder MapCatalogApi(this IEndpointRouteBuilder app
...
-                .ToListAsync();
+        List<CatalogItem> itemsOnPage = [];
+        var itemsWithDistance = services.CatalogAI.SearchMemoryAsync(text, pageSize);
 
-            services.Logger.LogDebug("Results from {text}: {results}", text, string.Join(", ", itemsWithDistance.Select(i => $"{i.Item.Name} => {i.Distance}")));
-
-            itemsOnPage = itemsWithDistance.Select(i => i.Item).ToList();
-        }
-        else
+        await foreach (var item in itemsWithDistance)
```
roji, line 139: Noting that in addition to losing pagination and the distance, this performs an additional database roundtrip for each result (so for a page size of 5, this does one query via SearchMemoryAsync, and then 5 FindAsync invocations totaling 6)... That's not great in terms of performance.

roji (commented):
@aaronpowell can I ask for a bit more context on the goal of this PR?

I think that generally, eShop is supposed to provide a sample of how an actual real-world app would be written (at least I think of it that way). If I were to write such an app, and have chosen pgvector as my vector store solution, I think I'd code directly against it - as the code currently is - rather than introduce an abstraction, which in this particular case seems to cause more friction than helping... Of course, the abstraction would 100% make sense in other scenarios, e.g. when using the vector DB within the SK stack (where it would be necessary), but I'm not sure it makes sense here...

What do you think?

PR #301, opened 2024-04-12
Refactor CatalogAI to support different embedding providers

stephentoub:
- CatalogAI is now implemented solely in terms of ITextEmbeddingGenerationService, rather than being OpenAI-specific
- Vector size has been reduced to 384. This is a more common size for local models, and the newer OpenAI text embedding models are sliceable down to this size while still retaining semantic meaning. In the future, we can allow the size to be customized in order to better accomodate more models.
- Catalog embeddings are now generated at startup using the provided embedding generator. Previously ones specific to the OpenAI embedding model were stored, but they were only valid for that one specific model. The generation has been refactored to generate them all in bulk rather than doing each in its own roundtrip to OpenAI. If using a local model, it's fast, regardless.


eerhardt:
Are the tests broken with this change? Or are they already broken in `main`?

src/Catalog.API/Catalog.API.csproj
```diff
@@ -21,6 +22,9 @@
   <!-- AI -->
   <ItemGroup>
     <PackageReference Include="Aspire.Azure.AI.OpenAI" />
+    <PackageReference Include="FastBertTokenizer" />
+    <PackageReference Include="Microsoft.SemanticKernel" />
+    <PackageReference Include="Microsoft.SemanticKernel.Connectors.Onnx" />
```
eerhardt, line 27: @adityamandaleeka and I were talking a bit about this last week. Is the intention of eshop to "be samples and show all possible approaches"? Or is it to be more like a "real app" and only do what someone would do in a "real app"?

Would someone generally have multiple ways of tokenizing in their app switchable by configuration? Or would they "code to abstractions", but only depend on one tokenizer implementation? Is the thought here that I would use `FastBertTokenizer` in local development, and then swtich to Open AI in production?

src/Catalog.API/Services/CatalogAI.cs
```diff
@@ -1,55 +1,76 @@
...
-        IsEnabled = _openAIClient != null;
+        _embeddingGenerator = embeddingGenerator;
         _environment = environment;
         _logger = logger;
+    }
+
+    /// <summary>Gets whether the AI system is enabled.</summary>
+    public bool IsEnabled => _embeddingGenerator is not null;
+
+    /// <summary>Gets an embedding vector for the specified catalog item.</summary>
```
eerhardt, line 27: (nit) Would it be better to have these "summary" docs on the interface and then just use `inheritdoc` here?

That way when in the IDE and using an `ICatalogAI` reference, the docs show up?

src/Catalog.API/Services/CatalogAI.cs
```diff
@@ -1,55 +1,76 @@
...
+    /// <summary>Gets embedding vectors for the specified catalog items.</summary>
+    public async ValueTask<IReadOnlyList<Vector>> GetEmbeddingsAsync(IEnumerable<CatalogItem> items)
+    {
+        if (IsEnabled)
         {
-            _logger.LogInformation("Embedding model: \"{model}\"", _aiEmbeddingModel);
+            long timestamp = Stopwatch.GetTimestamp();
+
+            IList<ReadOnlyMemory<float>> embeddings = await _embeddingGenerator.GenerateEmbeddingsAsync(items.Select(CatalogItemToString).ToList());
+            var results = embeddings.Select(m => new Vector(m[0..EmbeddingDimensions])).ToList();
```
eerhardt, line 41: Just for my knowledge, is there no way to pass `EmbeddingDimensions` into the embedding generator? That way it doesn't need to do extra work just to get discarded later.

src/Catalog.API/Catalog.API.csproj
stephentoub: > Would someone generally have multiple ways of tokenizing in their app switchable by configuration?

It's not about the tokenizers, in fact I should be able to remove this FastBertTokenizer package reference; it's just used by the Onnx package.

> Is the thought here that I would use FastBertTokenizer in local development, and then swtich to Open AI in production?

It's about the embedding generator, and there are multiple reasons you might want to switch between them. As you say, you might want to use a local embedding generator in development and switch to a hosted service for production deployment / scale. You might want to be able to failover to a different generator if a given service is down. You might want to be able to do A/B testing between multiple generators. You might want to use a paid service up until you hit a certain budget and then fall back to a local model. And so on. You might not configure it exactly as I've done here (there are various ways you might have multiple embedding generators and then pick between them at run time), but I think what's here is a reasonable approximation.

src/Catalog.API/Services/CatalogAI.cs
stephentoub: Sure

stephentoub:
Thanks, @eerhardt. Could you take another look?

stephentoub:
> Are the tests broken with this change? Or are they already broken in `main`?

They were already broken in main and have since been fixed in main, so this PR is now green as well.

src/Catalog.API/Extensions/Extensions.cs
```diff
@@ -27,12 +28,15 @@ public static void AddApplicationServices(this IHostApplicationBuilder builder)
...
-        if (!string.IsNullOrWhiteSpace(builder.Configuration.GetConnectionString("openai")))
+        if (builder.Configuration["AI:Onnx:EmbeddingModelPath"] is string modelPath &&
+            builder.Configuration["AI:Onnx:EmbeddingVocabPath"] is string vocabPath)
+        {
+            builder.Services.AddBertOnnxTextEmbeddingGeneration(modelPath, vocabPath);
+        }
+        else if (!string.IsNullOrWhiteSpace(builder.Configuration.GetConnectionString("openai")))
         {
             builder.AddAzureOpenAIClient("openai");
+            builder.Services.AddOpenAITextEmbeddingGeneration(builder.Configuration["AIOptions:OpenAI:EmbeddingName"] ?? "text-embedding-3-small");
```
eerhardt, line 39: Can you also change this in the AppHost so they are aligned?

https://github.com/dotnet/eShop/blob/d052bab01059e7e72bd0736fc63336d12a981551/src/eShop.AppHost/Program.cs#L77-L100

eerhardt (approved):
Looks good. Just one change to align some strings.

PR #308, opened 2024-04-15
Update OpenTelemetry packages

martincostello:
I pulled down the solution to look at #306, and the solution won't build due to NU1902 warnings.

Resolves GHSA-vh2m-22xx-q94f and #307 which I noticed someone opened just now.

PR #309, opened 2024-04-15
Use Polly.Core retry pipeline

### LINKED ISSUE #306: Use new Polly.Core APIs
I notice that the samples are still using the "old v7" Polly API to do retries:

https://github.com/dotnet/eShop/blob/705bb76734a32b9e7442ffe6f2041e38b8f26f12/src/EventBusRabbitMQ/RabbitMQEventBus.cs#L32-L34

Our new high-performance v8 API in Polly.Core would be better used for an example app (IMHO). Polly.Core is also AoT compatible.

Happy to do a PR if that's a welcome change.

martincostello:
- Use Polly.Core for retries.
- Fix IDE suggestions.
- Removes (seemingly) redundant reference to Polly.

Resolves #306.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -77,7 +77,7 @@
     <PackageVersion Include="FluentValidation.AspNetCore" Version="11.3.0" />
     <PackageVersion Include="MediatR" Version="12.2.0" />
     <PackageVersion Include="Microsoft.Web.LibraryManager.Build" Version="2.1.175" />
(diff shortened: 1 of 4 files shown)


martincostello:
Depends on #308.

adityamandaleeka (approved):

adityamandaleeka:
@sebastienros Can you take a quick look at this one too?

sebastienros (approved):

PR #311, opened 2024-04-16
Updates for Auth and Design

michaelstonis:
Updates For:

- Closer match for the new UI design
- Replace old authentication
- Fix auth token issues
- Remove old libraries like PCLCrypto
- UI performance changes
- Unit test cleanup


adityamandaleeka:
@michaelstonis Thanks for the PR. We'll need some time to review this one since it's so big, but we'll get back to you.

src/Identity.API/Program.cs
```diff
@@ -17,7 +17,7 @@
 
 builder.Services.AddIdentityServer(options =>
 {
-    options.IssuerUri = "null";
+    //options.IssuerUri = "null";
```
eerhardt, line 20: Why this change?

If we are going to make it, should the line just be deleted?

michaelstonis: Having this value set caused the mobile client to fail authentication. I have no problem removing it though.

michaelstonis:
@danmoseley I can, but a lot of it is interdependent on the .NET 8 upgrade. Let me take a pass at splitting it out into multiple.

Issue #298, opened 2024-04-11
[WebToolsE2E][Aspire] When running eShop.Web using http, the identity-api/webhooksclient project failed to start: Failed to apply configuration value 'WebhooksWebClient' / 'CallBackUrl'. A dependency may have failed to start.


v-sherryfan:
**INSTALL STEPS**

1. Clean machine: Win11 x64 23h2 ENU
2. Install VS 17.10 Preview 3 [ Includes Aspire 8.0.0-preview.5.24201.12 and SDK 8.0.300-preview.24203.14 ]
3. Install Docker Desktop for Windows

**REPRO STEPS**

1. Clone eShop repo, checkout the `aspire-preview.5` branch
2. Update sdk version to 8.0.300-preview.24203.14 in global.json
3. Open eShop.Web.slnf
4. Build 
5. F5 / Ctrl+F5 with **http**. Sometimes you will see this issue [#3435](https://github.com/dotnet/aspire/issues/3435), the workaround is: give it a minute, close the windows and try again - it works fine.

**ACTUAL**
![image](https://github.com/dotnet/eShop/assets/71361488/f7011425-1dc2-4a23-a0c7-e23ac31029e7)
![image](https://github.com/dotnet/eShop/assets/71361488/88d56cda-abe7-4a9e-a239-6bf7ddb054e7)
![image](https://github.com/dotnet/eShop/assets/71361488/17f89773-b067-4d6d-90fd-298dc19349c2)
![image](https://github.com/dotnet/eShop/assets/71361488/9967b0d0-4041-421b-a23d-af0449151e93)

**EXPECTED**

F5 / Ctrl+F5 works fine with https.
![image](https://github.com/dotnet/eShop/assets/71361488/7c9c2656-a1cf-4068-bc6f-40c20937fd94)

ChuanGz:
Hi @v-sherryfan, I think, for make sure protocol competible with all components, we can find all solution and make sure all run with http or https. Hope this help.

adityamandaleeka:
Can you please try this on main now that the preview5 PR has been merged along with some bugfixes/updates?

v-sherryfan:
@adityamandaleeka Yeah, I didn't reproduce the issue on the dotnet/eShop main branch either. So closed this issue as fixed.

dangerousor:
@v-sherryfan I got the same issue when running with `https` profile.

IEvangelist:
[...]

PR #322, opened 2024-04-22
fix: init webhooksClient with launchProfile

dangerousor:
Fix issue #298 
Init the `webhooksClient` with specific profile.

### DIFF
--- a/src/eShop.AppHost/Program.cs
+++ b/src/eShop.AppHost/Program.cs
@@ -57,7 +57,7 @@ builder.AddProject<Projects.Mobile_Bff_Shopping>("mobile-bff")
     .WithReference(identityApi);
 
 // Apps
-var webhooksClient = builder.AddProject<Projects.WebhookClient>("webhooksclient")
+var webhooksClient = builder.AddProject<Projects.WebhookClient>("webhooksclient", launchProfileName)
     .WithReference(webHooksApi)
     .WithEnvironment("IdentityUrl", identityEndpoint);
 


dangerousor:
@adityamandaleeka Can you help to give this an approval. I would say this can fix the issue #298.
image

mitchdenny:
I am wondering whether it is necessary to specify the launch profile here or not. When Aspire launches a child service project, that service project "inherits" the launch profile that was used to launch Aspire itself. If there is not a matching launch profile then it'll use the first launch profile specified in launch settings.

You can achieve a lot by aligning the launch profiles between the AppHost and the service projects.

dangerousor:
Wow, that's intereting. While as I tested on my machine, with Aspire preview5, it does not work to "inherits" automatically for the project `webhookclient`...

dangerousor:
@mitchdenny 
As we can see in the dashboard, for those no launch profile configured project, they all run with `http`, while in fact I launch the eShop with `http` profile.
So, basically, maybe it's better to explicitly launch all projects with specific profile, by setting the specific profile when `AddProject`?
For the issue #298, since the profile mismatch for the `CallbackUrl`, the eShop launch fails.

mitchdenny:
I would argue its better to have your launch profiles aligned (in terms of their name) across app host and service projects.

dangerousor:
OK...
Do you have any thoughts on the issues we are fixing? If we start the `eShop` with `https` profile, it should end up with passing the configuration to all the child projects. But seems something wrong. Can you repro it on your machine?

mitchdenny:
Can you still repro this issue? There was an issue with launch profiles that may have impacted you.

dangerousor:
I can try it later. I have Aspire preview 5 on my machine. Should I upgrade to the latest version of Aspire?

michaelstonis:
I am also hitting this issue with Aspire `8.0.0-preview.6.24214.1/8.0.100`. The fix applied here is the only way that I can successfully run HTTPS using macOS.

dangerousor:
Exactly I ran into this issue on my MacBook too...

cardinalip-williamkeller (approved):

PR #326, opened 2024-04-23
fix for issue 110

### LINKED ISSUE #110: Storing credit card information in the clear is a problem
Nice work on eshop/Aspire! I was looking through the code to get a better handle on Aspire and other things when I came across this migration:

https://github.com/dotnet-architecture/eShopOnAzure/blob/master/src/Services/Ordering/Ordering.API/Infrastructure/Migrations/20170208181933_Initial.cs#L91

It looks like a table is created in the database with columns to store card number, expiration date, and security code - all in the clear. Looking at the command/model for saving this information:

https://github.com/dotnet/eShop/blob/main/src/Ordering.API/Application/Commands/CreateOrderCommand.cs#L78

It doesn't look like there's any encryption or attempt to conceal private information for the card or the user. Running the app and placing an order confirms this:

screenshot_225

This is kind of a Big Deal when it comes to [PCI Compliance](https://www.nerdwallet.com/article/small-business/pci-compliance) which many companies are held to.  Moreover it's just a good idea to not store credit card information in the clear in a database :). Given that we refer to this as a "reference app" it might be a good idea to _do the needful_ for people so they don't accidentally do Bad Things.

## Recommendations

One idea that we could do here would be to call out, explicitly, in the API that we're not logging/storing the credit card information. At that point you could return a FakePayment object, for instance, with the information the app needs.

Another idea would be to stub out a Stripe Webhook Receiver with Stripe Checkout integration. Easy to do and I would be more than happy to PR that!

SrushtiPasari94:
Always creating the fake payment object with the same details regardless of what the user enters.

Fixes #110

### DIFF
--- a/src/WebApp/Components/Pages/Checkout/Checkout.razor
+++ b/src/WebApp/Components/Pages/Checkout/Checkout.razor
@@ -114,10 +114,8 @@
 
     protected override void OnInitialized()
     {
-        if (Info is null)
-        {
-            PopulateFormWithDefaultInfo();
-        }
+        //Always creating the fake payment object
+        PopulateFormWithDefaultInfo();
 
         editContext = new EditContext(Info!);
         extraMessages = new ValidationMessageStore(editContext);


jamesmontemagno (approved):

adityamandaleeka:
The issue here was about how the card info is being stored in the DB. Looks like this PR doesn't change anything about how that works.

The idea is to not store anything in the DB at all, and just pass in the fake info to the ordering API.

adityamandaleeka (changes requested):
See other comment

jamesmontemagno:
Ahh gotcha :) Thanks @adityamandaleeka  can you make these changes so it always sends the fake data regardless of what the user sends in. can just be 1111111111111111

adityamandaleeka:
> can you make these changes so it always sends the fake data regardless of what the user sends in. can just be 1111111111111111

I believe that was directed to @SrushtiPasari94 ?

jamesmontemagno:
I see now, just don't store anything at all for it. Talked to Srushti on this and she will update.

jamesmontemagno:
Talked to @SrushtiPasari94  and is going to update ASAP to simply not pass any data or store any data.

PR #340, opened 2024-05-03
Update to Aspire Preview 7

adityamandaleeka:


jamesmontemagno:
@adityamandaleeka with P7 out should we merge this?

adityamandaleeka:
We've been merging the ones from dependabot instead (it should open one soon). This is just for early testing.

danmoseley:
Note dependabot is flaky at the moment, so if we want something it can be worth kicking it to have another go. I just did.

jamesmontemagno:
I think all we need to do to this PR is just remove the nuget config changes... we should do that and merge this IMHO

danmoseley:
OK, let's try that.

adityamandaleeka:
I can't sign off on my own PR 😆.

jamesmontemagno (approved):

PR #359, opened 2024-05-08
Addressed the credit card number issue

### LINKED ISSUE #110: Storing credit card information in the clear is a problem

robconery:
The credit card number for an order was being stored in the database, saved to the user record as well as payment methods and orders. In addition, the card number was also logged in clear text.

I added a mask at the API request point for a new order, so instead of `4111111111111111` it is now `XXXXXXXXXXXX1111`. This is **not** an optimal solution, but is one step better than storing things in clear text.

Fixes #110


.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
+{
```
timheuer, line 1: I'd remove .vscode launch assets -- the pre-release/may release of the C# Dev Kit has a debugger launcher support for Aspire and children processes.

src/Ordering.API/Apis/OrdersApi.cs
```diff
@@ -135,9 +137,10 @@ public static async Task<OrderDraftDTO> CreateOrderDraftAsync(CreateOrderDraftCo
 
         using (services.Logger.BeginScope(new List<KeyValuePair<string, object>> { new("IdentifiedCommandId", requestId) }))
         {
+            var maskedCCNumber = request.CardNumber.Replace(request.CardNumber.Substring(0, 12), new string('X', 12));
```
timheuer, line 140: I have no idea what is more efficient for the compiler, but allocating a new string may not be? I would have used PadLeft but maybe someone tells me that's bad.


`request.CardNumber.Substring(request.CardNumber.Length - 4).PadLeft(request.CardNumber.Length, 'X');`

robconery: Given that it's an order, which isn't a high volume thing, is this really an issue?

src/Ordering.API/Apis/OrdersApi.cs
```diff
@@ -120,12 +120,14 @@ public static async Task<OrderDraftDTO> CreateOrderDraftAsync(CreateOrderDraftCo
...
+        
+        //mask the credit card number
+        
         services.Logger.LogInformation(
             "Sending command: {CommandName} - {IdProperty}: {CommandId} ({@Command})",
             request.GetGenericTypeName(),
             nameof(request.UserId),
-            request.UserId,
-            request);
+            request.UserId, new{}); //don't log the request as it has CC number
```
ReubenBond, line 130: Why not remove the `Command` parameter altogether?

robconery: I'm not too sure what you're asking; this code is a bit hard for me to follow and I went with "least impact" wherever I could.

ReubenBond:
Ideally, this would show people how to securely deal with payments and eliminate credit card storage altogether, rather than masking payment details without facilitating implementation of a secure payment flow.

I'm _ok_ with this for now, but it doesn't lead people in the right direction currently.

src/Ordering.API/Apis/OrdersApi.cs
```diff
@@ -120,12 +120,14 @@ public static async Task<OrderDraftDTO> CreateOrderDraftAsync(CreateOrderDraftCo
         CreateOrderRequest request,
         [AsParameters] OrderServices services)
     {
+        
+        //mask the credit card number
+        
         services.Logger.LogInformation(
             "Sending command: {CommandName} - {IdProperty}: {CommandId} ({@Command})",
```
ReubenBond, line 127:
```suggestion
            "Sending command: {CommandName} - {IdProperty}: {CommandId}",
```

ReubenBond: This, then delete the `, new{}` on line 130 below

src/Ordering.API/Apis/OrdersApi.cs
ReubenBond: I added a diff above to demonstrate

robconery:
I did start down that route (removing CC storage) but I wanted to stay focused on the task at hand, which is _triage_ for Build. Once we get this done, we can move on to a more ideal situation - but the CC bits are in the fabric of the application, meaning that the application requires a user to login, then we attach a payment method to them (which is CC rn), and that payment method ID is attached to the order. This is a weird way of doing eCommerce, but we can think on that when it's not on fire :D

ReubenBond:
This 2-step approach is fine to me. I see that CI is failing, eg the `eShop.Ordering.FunctionalTests.OrderingApiTests.AddNewOrder` test. Could you please fix the tests?

Any help you can provide in getting the payments flow up-to-scratch is much appreciated.

robconery:
Thanks James - should have had a null check in there!

adityamandaleeka:
Looks like this has already been merged, but I'd like to see a comment added here as well that indicates why we're doing the masking and explicitly point out that this isn't the "solution" to this problem. In the interest of guiding people down the right path, we wouldn't want people to come to this code and just remove the masking and move on...

PR #360, opened 2024-05-09
Update eShop to use built-in OpenAPI support

captainsafia:


src/eShop.ServiceDefaults/OpenApi.Extensions.cs
```diff
@@ -21,74 +21,114 @@ public static IApplicationBuilder UseDefaultOpenApi(this WebApplication app)
...
-        services.AddEndpointsApiExplorer();
-
         if (apiVersioning is not null)
         {
             // the default format will just be ApiVersion.ToString(); for example, 1.0.
             // this will format the version as "'v'major[.minor][-status]"
-            apiVersioning.AddApiExplorer(options => options.GroupNameFormat = "'v'VVV");
-            services.AddTransient<IConfigureOptions<SwaggerGenOptions>, ConfigureSwaggerOptions>();
-            services.AddSwaggerGen(options => options.OperationFilter<OpenApiDefaultValues>());
+            var versioned = apiVersioning.AddApiExplorer(options => options.GroupNameFormat = "'v'VVV");
```
captainsafia, line 106: @commonsensesoftware I think this does the trick WRT to applying the eager resolution pattern we talked about.

commonsensesoftware: @captainsafia mechanically, this looks good to me. You'll know that everything is working correctly if `IApiVersionDescriptionProvider` is picking up the collated set of API versions. To be sure, I would temporarily add (or change) an `Endpoint` to be something other than the default; for example, `2.0`. Double-check that it shows up. If it does, then you _should_ be good. When there are no collated API versions at all (which _shouldn't_ happen), the `IApiVersionDescriptionProvider` will always still include the value from `ApiVersioningOptions.DefaultApiVersion`, which is `ApiVersion.Default` (e.g. `1.0`) if otherwise unchanged. This can sometimes lead to a false sense of correctness if there is only one current API version defined and it happens to be `1.0`.

src/eShop.ServiceDefaults/OpenApiOptionsExtensions.cs
```diff
@@ -0,0 +1,229 @@
...
+    public static OpenApiOptions ApplyOperationDefaultValues(this OpenApiOptions options)
+    {
+        options.UseOperationTransformer((operation, context, cancellationToken) =>
+        {
+            var apiDescription = context.Description;
+
+            operation.Deprecated |= apiDescription.IsDeprecated();
+
+            // remove any assumed media types not present in the api description
+            foreach (var responseType in context.Description.SupportedResponseTypes)
```
commonsensesoftware, line 142: @captainsafia this _might_ not be needed anymore. this was a specific issue with Swashbuckle. I'm not sure if the problem still exists.

If you version by media type (the _truly_ RESTful way 😉), some incorrect assumptions were made. There are multiple possible approaches in an implementation, but the simplest is `application/json; v=#.#`. The API Explorer extensions use the defined `ApiDescription.SupportedRequestFormats`, `ApiDescription.SupportedResponseTypes`, and fans them out per API version. This will approach will take `application/json` and fan it out to `application/json; v=1.0`, `application/json; v=2.0`, and so on. If you're remotely curious how that happens, you can have a peek [here](https://github.com/dotnet/aspnet-api-versioning/blob/3fc071913dcded23eeb5ebe55bca44f3828488bf/src/AspNetCore/WebApi/src/Asp.Versioning.Mvc.ApiExplorer/ApiVersionParameterDescriptionContext.cs#L202).

This can be problematic because if _assuming_ is allowed (e.g. `AssumeDefaultVersionWhenUnspecified = true`), then `application/json` will work. If it's not, which is the default, then it will not work, but `application/json` is listed as a valid media type. While it's technical possible, there is no support for yielding just `application/json` if matching an implicit version is allowed. If the default version were `1.0`, then `application/json` and `application/json; v=1.0` are equivalent anyway. This is also true for an API _version-neutral_ endpoint.

If this behavior doesn't exist anymore, then TL;DR - I get it. This behavior does still exist, it should **only** happen if there are no other media types defined (IMO). I don't believe that behavior needs to exist here though.

captainsafia: Yeah, upon further inspection, we don't need to handle setting default values, fixing requiredness on parameters, or fixing parameters on descriptions.

src/eShop.ServiceDefaults/OpenApi.Extensions.cs
captainsafia: Yep, I was a little too trigger-happy to declare that this works. A more thorough validation that including an additional API version proves that this doesn't work.

When the enumerate the `IApiVersionDescriptionProvider` upfront at startup, the mapped API versions are not yet registered so the eager resolution on the builder doesn't catch these values. This strategy works for Swashbuckle because the document name doesn't play a key role in the resolution of OpenAPI-related services like it does in the built-in implementation.

There's a bit of a chicken-egg problem to unravel here. `AddOpenApi` needs to know all the version names ahead of time in order to configure the keyed service dependencies correctly. `Asp.Versioning` doesn't resolve the full set of supported versions until _after_ the endpoints are mapped.

An "ugly" workaround is to require that users explicitly map the supported version numbers themselves in their startup logic:

```csharp
using Asp.Versioning.Builder;

var builder = WebApplication.CreateBuilder(args);

builder.Services
	.AddOpenApi("v1")
	.AddOpenApi("v2");

var app = builder.Build();

app.MapOpenApi();

app.NewVersionedApi("Catalog")
   .MapCatalogApiV1()
   .MapCatalogApiV2();

app.Run();
```

captainsafia:
Gonna merge this into the release/9.0 branch since I'd like to work on updating that to consume the latest RC1 bits.

PR #372, opened 2024-05-14
remove card UI

danmoseley:
Minimal change to avoid collecting a credit card. eshop needs reworking to use a payment service.

I don't know the code at all - is this reasonable?


danmoseley:
If we do this we'll need a similar change in https://github.com/Azure-Samples/eShopOnAzure

jamesmontemagno:
@SrushtiPasari94 can you test this

SrushtiPasari94:
These changes are removing the Payment section form the Checkout Page. But otherwise working as expected.

image

@jamesmontemagno  Seems that is not expected to remove the Payment section.

danmoseley:
@SrushtiPasari94 I did intend to remove the payment section, since there is no longer a way to make payment.

Clearly, this isn't the long term solution. We likely want eshop to use some kind of payment service or a mock of one. I did try removing credit card entirely from eshop, but the changes are much larger and I don't have the time right now to complete that.

danmoseley:
What are your thoughts?

SrushtiPasari94:
@jamesmontemagno Please confirm if its ok to remove the Payment section?

@danmoseley The concern was that we don't want to log/store the credit card information anywhere in code. And that needs to be removed. **This PR is working fine** and shows in UI, that we are not getting information from user. But in code [CreateOrderCommand.cs](https://github.com/dotnet/eShop/blob/main/src/Ordering.API/Application/Commands/CreateOrderCommand.cs) we are storing credit card info in the DB.

Note: If this PR is just to remove the Payment section, it is working fine, and I can approve this.

danmoseley:
@SrushtiPasari94 By removing the UI to enter credit card info it's not possible for real credit card info to make its way into the database, right?

Yes it all needs ripping out, I was trying to make the minimal change for now.

@adityamandaleeka wdyt?

adityamandaleeka:
Looks fine to me as a temporary stopgap until we have a better example here.

PR #389, opened 2024-05-20
Moved Http playright test cases to use Https

SrushtiPasari94:


.github/workflows/playwright.yml
```diff
@@ -32,6 +32,11 @@ jobs:
         node-version: lts/*
     - name: Install dependencies
       run: npm ci
+    - name: Install .NET HTTPS Development Certificate
+      # if: matrix.os == 'ubuntu-latest'
+      run: |
+        dotnet dev-certs https --clean
```
davidfowl, line 38: dev certs don't work on linux. Did you test this? https://github.com/tmds/linux-dev-certs

SrushtiPasari94: Was just trying as linux cert was also not working in CI. Undoing this change

SrushtiPasari94: Used Linux cert

playwright.config.ts
```diff
@@ -22,7 +22,7 @@ export default defineConfig({
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
     /* Base URL to use in actions like `await page.goto('/')`. */
-    baseURL: 'http://localhost:5045',
+      baseURL: 'http://localhost:5045',
```
jamesmontemagno, line 25: @SrushtiPasari94 should we update this to https endpoing?

SrushtiPasari94: This also redirect to https.
Somehow with https playwright test are failing. So kept this as is and added middleware for httpsredirect.

jamesmontemagno: gotcha!

jamesmontemagno:
Updated and tested out mobile clients and seem to be working well here.

DamianEdwards (approved):
This looks like a good incremental improvement. We should look at making all the projects use HTTPS in a follow-up PR.

PR #392, opened 2024-05-23
Ai monitor updates

luisquintanilla:


davidfowl:
This version of eShop does not have azure dependencies. See https://github.com/Azure-Samples/eShopOnAzure

PR #406, opened 2024-06-05
Adds SK Telemetry using Semantic Conventions for AI

luisquintanilla:


luisquintanilla:
cc: @stephentoub

src/WebApp/Program.cs
```diff
@@ -1,12 +1,26 @@
...
+// Configure SK Telemetry
+builder.Services.AddOpenTelemetry()
+    .WithMetrics(m =>
+    {
+        m.AddMeter("Microsoft.SemanticKernel*");
+    })
+    .WithTracing(t =>
+    {
+        t.AddSource("Microsoft.SemanticKernel*");
+    });
```
stephentoub, line 22: Should all of these changes go into service defaults where the rest of telemetry is configured?

luisquintanilla: Good question. We can do it there as well, but AFAIK we're mainly surfacing the information from the web app. Not sure if we'd get additional info if it was baked into service defaults.

luisquintanilla:
Closing and reopening a new branch based on main.

PR #464, opened 2024-07-19
Add Commented Out Code for Azure Native Monitoring

mattmccleary:


src/eShop.AppHost/Program.cs
```diff
@@ -18,9 +18,14 @@
 
 var launchProfileName = ShouldUseHttpForEndpoints() ? "http" : "https";
 
+// To configure Azure Monitor Application Insights, add the following:
+// var insights = builder.AddConnectionString("myInsightsResource", "APPLICATIONINSIGHTS_CONNECTION_STRING");
```
samsp-msft, line 22: This can be replaced with a conditional such as 

``` c#
var appInsights = (builder.Configuration.GetConnectionString("myAppInsightsResource") is not null)? 
    builder.AddConnectionString(
        "myAppInsightsResource",
        "APPLICATIONINSIGHTS_CONNECTION_STRING")
    : null;
```

src/eShop.AppHost/Program.cs
```diff
@@ -18,9 +18,14 @@
...
+// You also need to add .WithReference(insights) to each service that should be monitored.
+
 // Services
 var identityApi = builder.AddProject<Projects.Identity_API>("identity-api", launchProfileName)
     .WithExternalHttpEndpoints()
+//  .WithReference(insights)    // Sample reference to Azure Monitor Application Insights. Copy/paste for each service.
```
samsp-msft, line 28: I was going to add an overload to have a conditional reference, but it already seems to be there, so the following could be added to each app and *not* need to be commented out. It should silently ignore it if the resource is null.

``` c#
    .WithReference(appInsights,null,true);
```

@mitchdenny - It turns out this doesn't work as the resource parameter isn't nullable. Is there a cleaner workaround?

src/eShop.AppHost/appsettings.json
```diff
@@ -7,6 +7,7 @@
     }
   },
   "ConnectionStrings": {
-    //"OpenAi": "Endpoint=xxxx;Key=xxxx"
+    // "OpenAi": "Endpoint=xxxx;Key=xxxx"
+    // "myInsightsResource": "REPLACE_WITH_YOUR_INSTRUMENTATION_KEY"  // Copy/paste from your Application Insights Resource in Azure Portal
```
samsp-msft, line 11: nope. This should be in user secrets.json, rather than here.

samsp-msft (commented):
I think it woudld be good to have this option in the app. Thoughts on how to make it cleaner?

src/eShop.AppHost/Program.cs
mattmccleary: Yeah the .WithReference doesn't like that appInsights could be null.
image

mitchdenny: Why not just use an conditional? I'm not sure we want to invent a new API for this?

samsp-msft: @mitchdenny conditionals don't fit nicely into the chained builder pattern, unless there is something I don't know?

mattmccleary:
@dotnet-policy-service agree company="Microsoft"

mattmccleary:
@samsp-msft Thoughts on my latest push? I added conditionals to each service reference.

samsp-msft:
> @samsp-msft Thoughts on my latest push? I added conditionals to each service reference.

If necessary, we can do that, but I really would prefer an ability to do a conditional cleanly in the apphost as its going to be a pretty common scenario.

@mitchdenny - How can we make conditionals clean in the apphost file?

samsp-msft:
We met and discussed the options - nothing is available right now that makes this cleaner. we should go in with the commented out blocks, and @mitchdenny  will file an issue for adding a conditional in the builder pattern.

mitchdenny:
Issue: https://github.com/dotnet/aspire/issues/5192

BrennanConroy:
Azure specific components are included in [eShopOnAzure](https://github.com/Azure-Samples/eShopOnAzure).

AppInsights (Azure Monitoring) is already included in eShopOnAzure: https://github.com/Azure-Samples/eShopOnAzure/blob/9da1eccec3454b19cd11a9b44001cd0323deb749/src/eShop.AppHost/Program.cs#L10

mattmccleary:
Great to see this on eShoponAzure!

I tested it, ran into a few issues.

1. Needed to update Aspire.Hosting.AppHost and Grpc.Tools packages to eliminate errors.
2. Needed to update Azure.Monitor.OpenTelemetry.AspNetCore for Live Metrics, a key feature
3. Several Projects basket-api, catalog-api, order-processor, ordering-api, payment-processor, and webapp, webhooks-api ran for a short amount of time then shows State "Finished". All other Projects, Containers, and Databases showed as "Running".

Also, I needed to add a Connection String to appsettings.json. Should we document that somewhere?

AnKushSingh05:
testing

PR #487, opened 2024-08-06
Support local .NET installation for eShop

captainsafia:
Adds support for a locally installed version of .NET for eShop to avoid polluting global .NET installations on dev machines.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -491,3 +491,6 @@ $RECYCLE.BIN/
 /playwright/.auth/
 /user.json
 .azure
+
+# Locally-installed .NET SDKs
+.dotnet/
--- a/README.md
+++ b/README.md
@@ -35,8 +35,29 @@ Or
 - From Dev Home go to `Machine Configuration -> Clone repositories`. Enter the URL for this repository. In the confirmation screen look for the section `Configuration File Detected` and click `Run File`.
 
 #### Mac, Linux, & Windows without Visual Studio
-- Install the latest [.NET 8 SDK](https://dot.net/download?cid=eshop)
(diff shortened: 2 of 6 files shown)


mitchdenny:
Given eshop is a sample is this something that we want others to replicate in their repos? If not I think we might optimising for our dev experience vs the educational value of the sample (as much as I would like these scripts).

captainsafia:
> Given eshop is a sample is this something that we want others to replicate in their repos? If not I think we might optimising for our dev experience vs the educational value of the sample (as much as I would like these scripts).

Good question! IMO,  I think there's two aspects to the pattern here:

The script to support automatically installing a given version of .NET given a global.json file. I don't think this is particularly unique to our setup given how popular the `dotnet-install.sh` scripts. I think this is fine to include in the sample.

The other aspect is the installation of the .NET SDK to a locally directory and the "activation" of that environment for a given session. This is the pattern that I suspect might be less common in "real world" codebases (?) but I don't necessarily think it is a bad pattern to encourage people to adopt.

mitchdenny:
Thats fair (about `dotnet-install`). I'm less bullish on having activate and friends but its a weakly held opinion (approved).

I'd argue that if we think this is genuinely useful, we should build something like this into the SDK itself ... some kind of "repo-local" install defined in `global.json`.

captainsafia:
In recent discussions about this, I realize it is sufficient to have this only in the `release/9.0` branch at the moment to support the work I am doing to get us ready for .NET 9 RC 1 which requires that we consume nightly SDK builds on a more frequent basis until we official ship RC1. So closing this for now.

PR #516, opened 2024-08-28
Remove Dapper Package Reference

bytefish:
Dapper is not used in the Ordering.API anymore, so the dependency should be removed.


bytefish:
@dotnet-policy-service agree

adityamandaleeka (approved):

adityamandaleeka:
Thanks @bytefish

PR #524, opened 2024-09-07
Upgrade Aspire.Hosting.AppHost to 8.2.0

Sean2xy:
To address the issue of "eShop.AppHost is a .NET Aspire AppHost project that needs a package reference to Aspire.Hosting.AppHo
st version 8.2.0 or above to work correctly" when first running the solution of eShop.AppHost.csproj, this PR bumps package of Aspire.Hosting.AppHost to 8.2.0.


adityamandaleeka:
AspireVersion was updated globally since then, so this PR can be closed.

PR #567, opened 2024-10-08
Update to .NET Aspire 9.0

eerhardt:


Directory.Packages.props
```diff
@@ -56,8 +56,10 @@
     <!-- Xabaril packages -->
     <PackageVersion Include="AspNetCore.HealthChecks.Uris" Version="8.0.1" />
     <!-- AI -->
-    <PackageVersion Include="Microsoft.SemanticKernel" Version="1.17.1" />
-    <PackageVersion Include="Microsoft.SemanticKernel.Connectors.Onnx" Version="1.14.0-alpha" />
+    <PackageVersion Include="Microsoft.Extensions.AI" Version="9.0.0-preview.9.24507.7" />
+    <PackageVersion Include="Microsoft.Extensions.AI.Abstractions" Version="9.0.0-preview.9.24507.7" />
+    <PackageVersion Include="Microsoft.Extensions.AI.OpenAI" Version="9.0.0-preview.9.24507.7" />
+    <PackageVersion Include="Microsoft.Extensions.AI.Ollama" Version="9.0.0-preview.9.24507.7" />
```
DamianEdwards, line 62: Can we not rely on transitive references for the first two here?

stephentoub: Possibly for Abstractions, but the latter two don't depend on the first.

src/Catalog.API/Extensions/Extensions.cs
```diff
@@ -28,17 +29,19 @@ public static void AddApplicationServices(this IHostApplicationBuilder builder)
         builder.Services.AddOptions<CatalogOptions>()
             .BindConfiguration(nameof(CatalogOptions));
 
-        if (builder.Configuration["AI:Onnx:EmbeddingModelPath"] is string modelPath &&
-            builder.Configuration["AI:Onnx:EmbeddingVocabPath"] is string vocabPath)
+        if (builder.Configuration["AI:Ollama:EmbeddingName"] is string ollamaEmbeddingModel)
         {
-            builder.Services.AddBertOnnxTextEmbeddingGeneration(modelPath, vocabPath);
+            builder.Services.AddSingleton<IEmbeddingGenerator<string, Embedding<float>>>(new OllamaEmbeddingGenerator(new Uri("http://localhost:11434"), ollamaEmbeddingModel));
```
DamianEdwards, line 34: What URL is this? Can we use service discovery or an injected address via configuration?

stephentoub: This is Ollama's default url. I had 10 minutes and just hardcoded it. The ideal would be we'd have an Ollama resource that would actually stand it up locally, get its url, feed it in, etc., but to my knowledge we don't right now. Second best would be pulling it from config, with a default that matches Ollama's default.

eerhardt: Do we need to support Ollama in eshop right now? We didn't support it before.

src/Catalog.API/appsettings.json
```diff
@@ -23,13 +23,14 @@
   },
   "CatalogOptions": {
     "UseCustomizationData": false
+  },
+  "AI": {
```
eerhardt, line 27: @stephentoub - did you intend to uncomment this block?

stephentoub: I did. Does it break something?

eerhardt: It doesn't appear to break anything. Using OpenAI still works for me without tweaking it.

src/eShop.AppHost/Program.cs
```diff
@@ -68,16 +72,16 @@
     .WithReference(basketApi)
     .WithReference(catalogApi)
     .WithReference(orderingApi)
-    .WithReference(rabbitMq)
+    .WithReference(rabbitMq).WaitFor(rabbitMq)
     .WithEnvironment("IdentityUrl", identityEndpoint);
 
 // set to true if you want to use OpenAI
-bool useOpenAI = false;
+bool useOpenAI = true;
```
eerhardt, line 79:
```suggestion
bool useOpenAI = false;
```

We should probably keep this off by default.

captainsafia:
I think we can also update the global.json file to point to the .NET 9 RC 2 SDK since it was released today?

maddymontaquila (approved):

nuget.config
```diff
@@ -4,6 +4,7 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+    <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
```
captainsafia, line 7: Nit: Note that this is needed for Aspire preview dependencies?

captainsafia (approved):
LGTM!

nuget.config
eerhardt: Yeah, I'm removing this now.

eerhardt:
This should be ready to merge once CI passes.

PR #568, opened 2024-10-09
Add OpenAPI metadata to Catalog API

mikekistler:
This PR adds OpenAPI metadata to the Catalog API.

It includes the generated OpenAPI doc, starting from the first commit, so that you can easily see how the doc was changed/improved by the app code changes.

I also included a Spectral ruleset file that checks basic structural correctness and adds rules to verify that all operations have both success and error responses defined and all parameters have descriptions.

The motivation for this work was so we could use this in demos of the OpenAPI document generation feature in upcoming conferences.

### DIFF
--- a/src/Catalog.API/Apis/CatalogApi.cs
+++ b/src/Catalog.API/Apis/CatalogApi.cs
@@ -1,5 +1,11 @@
 ﻿using System.ComponentModel;
+using System.Reflection;
+using Grpc.Core;
 using Microsoft.AspNetCore.Http.HttpResults;
+using Microsoft.AspNetCore.Http.Metadata;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Infrastructure;
+using Microsoft.VisualBasic;
 using Pgvector.EntityFrameworkCore;
 
 namespace eShop.Catalog.API;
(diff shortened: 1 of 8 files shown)


src/Catalog.API/Apis/CatalogApi.cs
```diff
@@ -1,4 +1,11 @@
-﻿using Microsoft.AspNetCore.Http.HttpResults;
+﻿using System.ComponentModel;
```
captainsafia, line 1: Do we need some of the usings here like `Grpc.Core` and `Microsoft.VisualBasic`?

src/Catalog.API/Apis/CatalogApi.cs
```diff
@@ -10,30 +17,87 @@ public static IEndpointRouteBuilder MapCatalogApiV1(this IEndpointRouteBuilder a
...
+        api.MapDelete("/items/{id:int}", DeleteItemById)
+            .WithName("DeleteItem")
+            .WithSummary("Delete catalog item")
+            .WithDescription("Delete the specified catalog item");
 
         return app;
     }
 
-    public static async Task<Results<Ok<PaginatedItems<CatalogItem>>, BadRequest<string>>> GetAllItems(
+    static DefaultProblemDetailsFactory problemDetailsFactory = new DefaultProblemDetailsFactory(null);
```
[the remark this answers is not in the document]
captainsafia: Edit: It looks like you'll want to do something like:

```
static DefaultProblemDetailsFactory problemDetailsFactory = new DefaultProblemDetailsFactory(Options.Create(new ApiBehaviorOptions()));
```

src/Catalog.API/Apis/CatalogApi.cs
```diff
@@ -52,21 +113,27 @@ public static async Task<Results<Ok<PaginatedItems<CatalogItem>>, BadRequest<str
...
         [AsParameters] CatalogServices services,
-        int id)
+        [Description("The catalog item id")] int id)
     {
         if (id <= 0)
         {
-            return TypedResults.BadRequest("Id is not valid.");
+            var problemDetails = problemDetailsFactory.CreateProblemDetails(httpContext,
+                detail: $"Id is not valid", statusCode: StatusCodes.Status400BadRequest);
+            httpContext.Response.ContentType = "application/problem+json";
```
mikekistler, line 135: Sadly, the `CreateProblemDetails` method fails to set the content-type to "application/problem+json". Even worse is that _despite_ adding the explicit "ProducesResponseType" on this method with the correct content-type, this doesn't make it into the OpenAPI document -- probably because the BadRequest<ProblemDetails> overrides it (incorrectly IMHO).

mikekistler:
Closing this PR as #631 is a better solution.

PR #581, opened 2024-10-18
Update .NET MAUI app to .NET 9

jamesmontemagno:
#### PR Classification
Code cleanup and upgrade to .NET 9.0.

#### PR Summary
This pull request updates the project to target .NET 9.0, includes code cleanup, and updates package references.
- `App.xaml.cs`: Overridden `CreateWindow` method to return a new `Window` instance with `AppShell` and updated `SetStatusBar` method.
- `ClientApp.csproj` and `HybridApp.csproj`: Target .NET 9.0 and updated package references.
- `Basket.cs`: Improved protobuf parsing by handling end group tags.
- `DialogService.cs`: Updated to use `AppShell.Current.DisplayAlert` instead of `Application.Current.MainPage.DisplayAlert`.
- `SettingsViewModel.cs`: Made `TitleUseAzureServices` property non-static.


src/ClientApp/App.xaml.cs
```diff
@@ -86,7 +90,7 @@ private void App_RequestedThemeChanged(object sender, AppThemeChangedEventArgs e
 
     private void SetStatusBar()
     {
-        var nav = Current.MainPage as NavigationPage;
+        var nav = Windows[0].Page as NavigationPage;
```
jamesmontemagno, line 93: @jfversluis  thoughts?

jamesmontemagno:
@jfversluis  can you pull this and take a look as to why teh fontawesome icons aren't loading....

src/HybridApp/App.xaml.cs
```diff
@@ -5,7 +5,10 @@ public partial class App : Application
 	public App()
 	{
 		InitializeComponent();
-
-		MainPage = new MainPage();
 	}
+
+    protected override Window CreateWindow(IActivationState? activationState)
+    {
+        return new Window(new MainPage());
```
Eilon, line 12: Let's make it even snazzier and have a title:

```suggestion
        return new Window(new MainPage()) { Title = "eShop" };
```

Apparently this has been a bug in the .NET MAUI Blazor Hybrid templates for only... 3 years... and we just fixed it 😁

Eilon (changes requested):

Eilon (approved):
I focused my review on the HybridApp, which I'm most familiar with, and I think it's good. I scanned through the ClientApp but only quickly and didn't find any issues.

src/ClientApp/App.xaml.cs
jfversluis: Yep that is unfortunately what it is for now. Alternatively since you're using Shell you could do `Shell.Current.CurrentPage` which is a tiny bit nicer? But also I think that might be going away at some point since it's also not great.

Additionally, I don't think the whole StatusBar thing does much anyway? 😅

https://github.com/dotnet/eShop/blob/main/src/ClientApp/Services/Theme/Theme.shared.cs#L5

jfversluis:
> @jfversluis can you pull this and take a look as to why teh fontawesome icons aren't loading....

Tested on iOS, macOS and Android and it shows the icons there, so maybe just a fluke?

jfversluis:
I do see a couple of 

> Microsoft.Maui.Controls.BindableObject: Warning: Cannot convert 6 to type 'System.Int32'

In the debug output, but can't really pinpoint where its coming from. Does that ring a bell?

jamesmontemagno:
was able to fix up the column span thing... nto srue about the "6" that is weird

It has to be somethign on the login page...

if i delete

```
 <Button
      Grid.Row="2"
      Grid.Column="0"
      Margin="8"
      BackgroundColor="Black"
      Command="{Binding SignInCommand}"
      CornerRadius="16"
      Text="Login"
      TextColor="White" />

```
the warnings go away

PR #587, opened 2024-10-22
Use source-generated JSON serialization in .NET MAUI app

simonrozsival:
Using source-generated JSON serialization is a necessary step to add support for NativeAOT at some later point.

/cc @jamesmontemagno


jamesmontemagno:
NOICE!

jamesmontemagno:
I changed the base for the .NET 9 stuff

PR #588, opened 2024-10-24
Update Scalar API reference config to avoid loading external assets

captainsafia:
Use new options in the Scalar API reference to avoid loading external fonts and to fetch the `scalar.js` dependency from the local NuGet package instead of a public CDN.

### DIFF
--- a/src/eShop.ServiceDefaults/OpenApi.Extensions.cs
+++ b/src/eShop.ServiceDefaults/OpenApi.Extensions.cs
@@ -27,8 +27,6 @@ public static partial class Extensions
         {
             app.MapScalarApiReference(options =>
             {
-                // Empty CDN url to support loading JavaScript assets from NuGet package
-                options.CdnUrl = string.Empty;
                 // Disable default fonts to avoid download unnecessary fonts
                 options.DefaultFonts = false;
             });
--- a/src/eShop.ServiceDefaults/OpenApi.Extensions.cs
+++ b/src/eShop.ServiceDefaults/OpenApi.Extensions.cs
@@ -1,5 +1,4 @@
 ﻿using Asp.Versioning;
-using Asp.Versioning.ApiExplorer;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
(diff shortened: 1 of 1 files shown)


src/eShop.ServiceDefaults/OpenApi.Extensions.cs
```diff
@@ -26,7 +25,13 @@ public static IApplicationBuilder UseDefaultOpenApi(this WebApplication app)
 
         if (app.Environment.IsDevelopment())
         {
-            app.MapScalarApiReference();
+            app.MapScalarApiReference(options =>
+            {
+                // Empty CDN url to support loading JavaScript assets from NuGet package
+                options.CdnUrl = string.Empty;
```
xC0dex, line 31: Please don’t ask how I stumbled upon this PR 😄 
Since `CdnUrl` is now `null` by default, setting it to `string.Empty` isn’t necessary.

captainsafia: Oooh! I hadn't realized this had changed. Good to know -- avoiding the CDN is nice default behavior.

src/eShop.ServiceDefaults/OpenApi.Extensions.cs
```diff
@@ -64,6 +67,13 @@ public static IHostApplicationBuilder AddDefaultOpenApi(
                     options.ApplyAuthorizationChecks([.. scopes.Keys]);
                     options.ApplySecuritySchemeDefinitions();
                     options.ApplyOperationDeprecatedStatus();
+                    // Clear out the default servers so we can fallback to
+                    // whatever ports have been allocated for the service by Aspire
```
BrennanConroy, line 71: What fallback exactly? I assume `Servers` won't be populated by anyone else?

captainsafia: It would be up to the consumer of the OpenAPI document to set it. When you're in Scalar, it will use the host that the Scalar UI is served on as the base host name. If you import the OpenAPI document into a tool like Bruno or Postman, they have UIs where you can set the base host name. This is a result of this issue: https://github.com/dotnet/aspnetcore/issues/57332

BrennanConroy (approved):

PR #625, opened 2024-11-18
Adding Ollama Aspire integration

aaronpowell:
Using the Ollama hosting integration from the Community Toolkit in the app host. There is a conditional in place that will not enable ollama support by default.

Using the OllamaSharp client integration to provide the IChatClient and IEmbeddingGenerator implementations.


src/eShop.AppHost/Program.cs
```diff
@@ -81,6 +81,24 @@
...
+        .WithOpenWebUI();
+    var embeddings = ollama.AddModel("embedding", "all-minilm");
+    var chat = ollama.AddModel("chat", "llama3.1");
+
+    catalogApi.WithReference(embeddings)
+        .WithEnvironment("OllamaEnabled", "true")
+        .WaitFor(embeddings);
+    webApp.WithReference(chat)
+        .WithEnvironment("OllamaEnabled", "true")
+        .WaitFor(chat);
```
eerhardt, line 99: Can we put this in Extensions.cs like the `AddOpenAI` method is?

src/WebApp/Extensions/Extensions.cs
```diff
@@ -98,14 +98,13 @@ public static void AddAuthenticationServices(this IHostApplicationBuilder builde
 
     private static void AddAIServices(this IHostApplicationBuilder builder)
     {
-        string? ollamaEndpoint = builder.Configuration["AI:Ollama:Endpoint"];
-        if (!string.IsNullOrWhiteSpace(ollamaEndpoint))
+        if (builder.Configuration["OllamaEnabled"] is string ollamaEnabled && bool.Parse(ollamaEnabled))
         {
-            builder.Services.AddChatClient(new OllamaChatClient(ollamaEndpoint, builder.Configuration["AI:Ollama:ChatModel"] ?? "llama3.1"))
+            builder.AddKeyedOllamaSharpChatClient("chat");
```
eerhardt, line 103: Why `keyed`?

src/eShop.AppHost/Program.cs
aaronpowell: Done.

src/WebApp/Extensions/Extensions.cs
aaronpowell: Hold over from a bug that I had where registrations weren't happening correctly in the Toolkit. That bug is fixed in 9.1 of the Toolkit so I've dropped the keyed service here (and in the catalog API)

Directory.Packages.props
```diff
@@ -97,6 +99,6 @@
     <PackageVersion Include="xunit.runner.visualstudio" Version="2.8.2" />
     <PackageVersion Include="Yarp.ReverseProxy" Version="2.2.0" />
     <PackageVersion Include="IdentityModel" Version="7.0.0" />
-    <PackageVersion Include="Scalar.AspNetCore" Version="1.2.44"/>
+    <PackageVersion Include="Scalar.AspNetCore" Version="1.2.36" />
```
eerhardt, line 102:
```suggestion
    <PackageVersion Include="Scalar.AspNetCore" Version="1.2.44" />
```

This is undoing a recent depend-a-bot update and it looks unrelated. Can we revert this change?

src/eShop.AppHost/Extensions.cs
```diff
@@ -75,4 +75,28 @@ public static IDistributedApplicationBuilder AddOpenAI(this IDistributedApplicat
...
+    /// Configures eShop projects to use Ollama for text embedding and chat.
+    /// </summary>
+    public static IDistributedApplicationBuilder AddOllama(this IDistributedApplicationBuilder builder,
+        IResourceBuilder<ProjectResource> catalogApi,
+        IResourceBuilder<ProjectResource> webApp)
+    {
+        var ollama = builder.AddOllama("ollama")
+            .WithDataVolume()
+            .WithGPUSupport()
+            .WithOpenWebUI();
```
eerhardt, line 89: How long does it take for the Ollama container to spin up? Should this have persistent lifetime?

src/Catalog.API/Extensions/Extensions.cs
```diff
@@ -37,12 +37,12 @@ public static void AddApplicationServices(this IHostApplicationBuilder builder)
         builder.Services.AddOptions<CatalogOptions>()
             .BindConfiguration(nameof(CatalogOptions));
 
-        if (builder.Configuration["AI:Ollama:Endpoint"] is string ollamaEndpoint && !string.IsNullOrWhiteSpace(ollamaEndpoint))
+        if (builder.Configuration["OllamaEnabled"] is string ollamaEnabled && bool.Parse(ollamaEnabled))
```
eerhardt, line 40: Since we've changed what configuration values get read here, can we also update the appsettings.json files for these changes?

https://github.com/dotnet/eShop/blob/633dd1a6525e705a85ed3f65e5167c2f901e51a7/src/WebApp/appsettings.json#L13-L20

https://github.com/dotnet/eShop/blob/633dd1a6525e705a85ed3f65e5167c2f901e51a7/src/Catalog.API/appsettings.json#L27-L34

eerhardt (commented):
This is looking pretty nice @aaronpowell!

I just had a few more comments. I think once those are addressed this should be good to merge.

src/eShop.AppHost/Extensions.cs
aaronpowell: Ollama itself is a pretty lightweight container and starts fast (I have it running on a NUC), so I wouldn't do a persistent lifetime. The expensive part is model downloading, which is why I added the `WithDataVolume` so that you only download the models on first run.

src/Catalog.API/Extensions/Extensions.cs
aaronpowell: ah yep, didn't think to dig into the appsettings to look for things that needed cleaning up

eerhardt (approved):
This looks great! Thanks for the contribution.

PR #631, opened 2024-11-24
Add OpenAPI metadata to Catalog API endpoints

mikekistler:
This PR adds summaries, descriptions, and tags to the Catalog API endpoints and descriptions on all parameters and properties.

### DIFF
--- a/src/Catalog.API/Apis/CatalogApi.cs
+++ b/src/Catalog.API/Apis/CatalogApi.cs
@@ -1,5 +1,11 @@
 ﻿using System.ComponentModel;
+using System.Reflection;
+using Grpc.Core;
 using Microsoft.AspNetCore.Http.HttpResults;
+using Microsoft.AspNetCore.Http.Metadata;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Infrastructure;
+using Microsoft.VisualBasic;
 using Pgvector.EntityFrameworkCore;
 
 namespace eShop.Catalog.API;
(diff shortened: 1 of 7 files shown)


jamesmontemagno (approved):

src/Catalog.API/Apis/CatalogApi.cs
```diff
@@ -10,30 +13,85 @@ public static IEndpointRouteBuilder MapCatalogApiV1(this IEndpointRouteBuilder a
...
+            .WithSummary("Get catalog items by type and brand")
+            .WithDescription("Get catalog items of the specified type and brand")
+            .WithTags("Types");
+        api.MapGet("/items/type/all/brand/{brandId:int?}", GetItemsByBrandId)
+            .WithName("GetItemsByBrand")
+            .WithSummary("List catalog items by brand")
+            .WithDescription("Get a list of catalog items for the specified brand")
+            .WithTags("Brands");
+        api.MapGet("/catalogtypes",
+            [ProducesResponseType<ProblemDetails>(StatusCodes.Status400BadRequest, "application/problem+json")]
```
captainsafia, line 61: Remind me again why we're not using `.ProducesProblem(StatusCodes.Status400BadRequest, "application/problem+json")`?

src/Catalog.API/Model/PaginationRequest.cs
```diff
@@ -1,3 +1,13 @@
-﻿namespace eShop.Catalog.API.Model;
+﻿using System.ComponentModel;
 
-public record PaginationRequest(int PageSize = 10, int PageIndex = 0);
+namespace eShop.Catalog.API.Model;
+
+public record PaginationRequest(
+    [property: Description("Number of items to return in a single page of results")]
+    [property: DefaultValue(10)]
+    int PageSize = 10,
```
captainsafia, line 8: Hmmm....we should probably support default parameter values instead of relying on the attribute as well. 🤔

captainsafia (approved):
Looks good overall! Had a question inline to jog my memory about why we're not using `ProducesProblem`.

PR #632, opened 2024-11-24
Add build-time OpenAPI document generation

mikekistler:
This PR adds build-time OpenAPI document generation for the Catalog service.

### DIFF
--- a/src/Catalog.API/Model/CatalogServices.cs
+++ b/src/Catalog.API/Model/CatalogServices.cs
@@ -1,11 +1,12 @@
 ﻿using eShop.Catalog.API.Services;
+using Microsoft.AspNetCore.Mvc;
 
 public class CatalogServices(
     CatalogContext context,
-    ICatalogAI catalogAI,
+    [FromServices] ICatalogAI catalogAI,
     IOptions<CatalogOptions> options,
     ILogger<CatalogServices> logger,
-    ICatalogIntegrationEventService eventService)
+    [FromServices] ICatalogIntegrationEventService eventService)
 {
     public CatalogContext Context { get; } = context;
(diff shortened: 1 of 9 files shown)


src/Catalog.API/Model/CatalogServices.cs
```diff
@@ -1,11 +1,12 @@
 ﻿using eShop.Catalog.API.Services;
+using Microsoft.AspNetCore.Mvc;
 
 public class CatalogServices(
     CatalogContext context,
-    ICatalogAI catalogAI,
+    [FromServices] ICatalogAI catalogAI,
```
captainsafia, line 6: Why is an explicit `[FromServices]` needed here?

mikekistler: Because at build time we bypass registering the services and then model-binding doesn't know what do to with these parameters. An alternative was to register dummy instances of these services at build time but this seemed like the cleaner approach.

jamesmontemagno (approved):

src/Catalog.API/Program.cs
```diff
@@ -1,9 +1,18 @@
 ﻿using Asp.Versioning.Builder;
+using System.Reflection;
 
 var builder = WebApplication.CreateBuilder(args);
 
-builder.AddServiceDefaults();
-builder.AddApplicationServices();
+if (Assembly.GetEntryAssembly()?.GetName().Name == "GetDocument.Insider")
```
captainsafia, line 6: Instead of putting this in `Program.cs` should we stick it in a helper method in the service defaults project?

If not, maybe we can stick this check into a nicer method name like `IsBuildTimeOpenApi`? Here's what I need in our benchmark apps at https://github.com/aspnet/Benchmarks/pull/2011.

PR #659, opened 2024-12-13
Refactor Catalog API

### LINKED ISSUE #589: Refactor Catalog API
The Catalog API should be refactored to comply with standard HTTP/REST conventions. Specifically:

- The PUT operation to update a catalog item should take the item id in the path
- The GET operations on
    - /api/catalog/items/by/{name}
    - /api/catalog/items/type/{typeId}/brand/{brandId}
    - /api/catalog/items/type/all/brand/{brandId}
-  should be dropped and replaced with optional query parameters `name`, `type`, and `brand` on the GET on /api/catalog/items
- The "text" parameter of GET on /api/catalog/items/withsemanticrelevance/{text} should be changed from a path parameter to a query parameter

mikekistler:
This PR refactors the Catalog API to adopt more commonly accepted REST API conventions.

Fixes #589.

### DIFF
--- a/src/Catalog.API/Apis/CatalogApi.cs
+++ b/src/Catalog.API/Apis/CatalogApi.cs
@@ -1,4 +1,5 @@
 ﻿using System.ComponentModel;
+using System.ComponentModel.DataAnnotations;
 using Microsoft.AspNetCore.Http.HttpResults;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
@@ -35,7 +36,7 @@ public static class CatalogApi
             .WithTags("Items");
 
         // Routes for resolving catalog items using AI.
-        api.MapGet("/items/withsemanticrelevance/{text:minlength(1)}", GetItemsBySemanticRelevance)
+        api.MapGet("/items/withsemanticrelevance", GetItemsBySemanticRelevance)
             .WithName("GetRelevantItems")
             .WithSummary("Search catalog for relevant items")
             .WithDescription("Search the catalog for items related to the specified text")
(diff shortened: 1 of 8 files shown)


src/Catalog.API/Apis/CatalogApi.cs
```diff
@@ -73,15 +58,14 @@ public static IEndpointRouteBuilder MapCatalogApiV1(this IEndpointRouteBuilder a
...
-        api.MapPut("/items", UpdateItem)
-            .WithName("UpdateItem")
-            .WithSummary("Create or replace a catalog item")
-            .WithDescription("Create or replace a catalog item")
-            .WithTags("Items");
         api.MapPost("/items", CreateItem)
             .WithName("CreateItem")
             .WithSummary("Create a catalog item")
             .WithDescription("Create a new item in the catalog");
+        api.MapPut("/items/{id:int}", UpdateItem)
```
captainsafia, line 65: Do we want to retain the `Items` tag on this and the other items-related endpoints?

src/Catalog.API/Apis/CatalogApi.cs
```diff
@@ -93,16 +77,33 @@ public static IEndpointRouteBuilder MapCatalogApiV1(this IEndpointRouteBuilder a
...
+        [AsParameters] CatalogServices services,
+        [Description("The name of the item to return")] string name,
+        [Description("The type of items to return")] int? type,
+        [Description("The brand of items to return")] int? brand)
     {
         var pageSize = paginationRequest.PageSize;
         var pageIndex = paginationRequest.PageIndex;
 
-        var totalItems = await services.Context.CatalogItems
+        var root = (IQueryable<CatalogItem>)services.Context.CatalogItems;
```
captainsafia, line 88: Do we need the cast to `IQueryable<CatalogItem>` here?

src/Catalog.API/Apis/CatalogApi.cs
```diff
@@ -93,16 +77,33 @@ public static IEndpointRouteBuilder MapCatalogApiV1(this IEndpointRouteBuilder a
...
+        {
+            root = root.Where(c => c.CatalogBrandId == brand);
+        }
+
+        var totalItems = await root
             .LongCountAsync();
 
-        var itemsOnPage = await services.Context.CatalogItems
-            .OrderBy(c => c.Name)
+        var itemsOnPage = await root
```
captainsafia, line 106: Looks like the ordering-by-name as removed here. Do we want to keep that?

src/Mobile.Bff.Shopping/appsettings.json
```diff
@@ -64,64 +64,10 @@
           }
         ]
       },
-      "route4": {
```
captainsafia, line 67: Should we keep the api-version query parameter here?

captainsafia (commented):
Left some comments inline!

Do we want to create a new API version for the catalog API and make these changes in a new version of this API? It would be a good way to dogfood the versioning experience.

src/Catalog.API/Apis/CatalogApi.cs
mikekistler: Yes. Without it `root` is `DbSet<CatalogItem>` and this gives build errors on the `Where` methods below.

mikekistler:
Closing in favor of #680

PR #680, opened 2025-01-07
Refactor Catalog API in new API version

### LINKED ISSUE #589: Refactor Catalog API

mikekistler:
This PR refactors the Catalog API to adopt more commonly accepted REST API conventions. The existing V1 API is unchanged and all refactoring is done in a new API version, V2.

Fixes #589.

Note: I have not tested the clients yet, so will move this to draft until that is done.

### DIFF
--- a/src/Catalog.API/Apis/CatalogApi.cs
+++ b/src/Catalog.API/Apis/CatalogApi.cs
@@ -42,7 +42,7 @@ public static class CatalogApi
             .WithSummary("Get catalog item")
             .WithDescription("Get an item from the catalog")
             .WithTags("Items");
-        api.MapGet("/items/by/{name:minlength(1)}", GetItemsByName)
+        v1.MapGet("/items/by/{name:minlength(1)}", GetItemsByName)
             .WithName("GetItemsByName")
             .WithSummary("Get catalog items by name")
             .WithDescription("Get a paginated list of catalog items with the specified name.")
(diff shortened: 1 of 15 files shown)


mikekistler:
I have tested the Blazor, Maui, and Hybrid App and all seem to be functioning correctly.

Moved to Ready for Review.

src/Catalog.API/Apis/CatalogApi.cs
```diff
@@ -8,12 +9,26 @@ namespace eShop.Catalog.API;
...
+
+        var vApi = app.NewVersionedApi("Catalog");
+        var api = vApi.MapGroup("api/catalog").HasApiVersion(1, 0).HasApiVersion(2, 0);
+        var v1 = vApi.MapGroup("api/catalog").HasApiVersion(1, 0);
+        var v2 = vApi.MapGroup("api/catalog").HasApiVersion(2, 0);
 
         // Routes for querying catalog items.
-        api.MapGet("/items", GetAllItems)
+        v1.MapGet("/items", GetAllItemsV1)
+            // .WithName("ListItems")
```
captainsafia, line 27: Maybe we can try the following for now?

```suggestion
            .WithName("ListItems-v1")
```

I think in an ideal world, there is some sort of `WithVersionedName` extension made that will generate a globally unique endpoint name by prepending the API version somewhere but this might be a good stop-gap.

src/Catalog.API/Catalog.API.json
captainsafia, line 1: I haven't looked super closely at the delta here but should we expect any changes in the V1 API at all? Ideally there would no changes to this document and only changes in V2, correct?

tests/Catalog.FunctionalTests/CatalogApiTests.cs
```diff
@@ -4,26 +4,33 @@
 using Asp.Versioning.Http;
 using eShop.Catalog.API.Model;
 using Microsoft.AspNetCore.Mvc.Testing;
+using Microsoft.EntityFrameworkCore.Query.SqlExpressions;
```
captainsafia, line 7: Is this used anywhere?

src/Catalog.API/Catalog.API.json
mikekistler: Yes that is what we should expect. And that is almost true except for two things:
1. The `UpdateItem` method (Put) now generates a `400 Bad Request` when the body does not contain a `id` property. This actually an improvement to the API as previously this was reported as a `404 Not Found`.
2. The OpenAPI generated by .NET (currently) does not have a deterministic structure, so the vast majority of the diffed lines here are just arbitrary reorderings. I opened dotnet/aspnetcore#59809 to track this.

captainsafia (approved):

PR #744, opened 2025-02-20
Optimize user creation for asynchronous calls

toyzoz:


danmoseley (approved):
thanks! this seems reasonable

PR #759, opened 2025-02-24
Upgrade to .NET Aspire 9.1

eerhardt:
- Update all Aspire package versions, including the latest .NET and ASP.NET Core versions.
- Remove LangVersion preview to work around https://github.com/npgsql/efcore.pg/issues/3461
- Update to latest MEAI
- Add support for using an existing OpenAI account via AsExisting

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,6 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion>preview</LangVersion>
     <WarnOnPackingNonPackableProject>false</WarnOnPackingNonPackableProject>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
--- a/nuget.config
+++ b/nuget.config
@@ -4,12 +4,16 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+    <add key="darc-pub-dotnet-aspire-2a8f48ea" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-aspire-2a8f48ea/nuget/v3/index.json" />
   </packageSources>
 
   <packageSourceMapping>
     <!-- key value for <packageSource> should match key values from <packageSources> element -->
     <clear />
     <packageSource key="nuget">
+      <package pattern="*" />
+    </packageSource>
+    <packageSource key="darc-pub-dotnet-aspire-2a8f48ea">
       <package pattern="*" />
     </packageSource>
   </packageSourceMapping>
--- a/nuget.config
+++ b/nuget.config
@@ -4,16 +4,12 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="nuget" value="https://api.nuget.org/v3/index.json" />
-    <add key="darc-pub-dotnet-aspire-2a8f48ea" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-aspire-2a8f48ea/nuget/v3/index.json" />
   </packageSources>
 
   <packageSourceMapping>
     <!-- key value for <packageSource> should match key values from <packageSources> element -->
     <clear />
     <packageSource key="nuget">
-      <package pattern="*" />
-    </packageSource>
-    <packageSource key="darc-pub-dotnet-aspire-2a8f48ea">
       <package pattern="*" />
     </packageSource>
   </packageSourceMapping>
--- a/src/Catalog.API/Extensions/Extensions.cs
+++ b/src/Catalog.API/Extensions/Extensions.cs
@@ -1,5 +1,6 @@
 ﻿using eShop.Catalog.API.Services;
 using Microsoft.Extensions.AI;
+using OllamaSharp;
 using OpenAI;
 
 public static class Extensions
@@ -40,7 +41,7 @@ public static class Extensions
         if (builder.Configuration["OllamaEnabled"] is string ollamaEnabled && bool.Parse(ollamaEnabled))
         {
             builder.AddOllamaSharpEmbeddingGenerator("embedding");
-            builder.Services.AddEmbeddingGenerator(b => b.GetRequiredService<IEmbeddingGenerator<string, Embedding<float>>>())
+            builder.Services.AddEmbeddingGenerator(sp => (IEmbeddingGenerator<string, Embedding<float>>)sp.GetRequiredKeyedService<IOllamaApiClient>("embedding"))
                 .UseOpenTelemetry()
                 .UseLogging();
         }
@@ -49,8 +50,7 @@ public static class Extensions
             builder.AddOpenAIClientFromConfiguration("openai");
             builder.Services.AddEmbeddingGenerator(sp => sp.GetRequiredService<OpenAIClient>().AsEmbeddingGenerator(builder.Configuration["AI:OpenAI:EmbeddingModel"]!))
                 .UseOpenTelemetry()
-                .UseLogging()
-                .Build();
+                .UseLogging();
         }
 
         builder.Services.AddScoped<ICatalogAI, CatalogAI>();
--- a/src/WebApp/Extensions/Extensions.cs
+++ b/src/WebApp/Extensions/Extensions.cs
@@ -1,16 +1,14 @@
-﻿using System;
-using Azure.AI.OpenAI;
+﻿using eShop.Basket.API.Grpc;
 using eShop.WebApp;
+using eShop.WebApp.Services.OrderStatus.IntegrationEvents;
 using eShop.WebAppComponents.Services;
 using Microsoft.AspNetCore.Authentication.Cookies;
 using Microsoft.AspNetCore.Authentication.OpenIdConnect;
 using Microsoft.AspNetCore.Components.Authorization;
 using Microsoft.AspNetCore.Components.Server;
 using Microsoft.Extensions.AI;
-using Microsoft.Extensions.DependencyInjection;
 using Microsoft.IdentityModel.JsonWebTokens;
-using eShop.WebApp.Services.OrderStatus.IntegrationEvents;
-using eShop.Basket.API.Grpc;
+using OllamaSharp;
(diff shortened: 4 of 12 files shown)


Directory.Build.props
```diff
@@ -1,7 +1,6 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion>preview</LangVersion>
```
eerhardt, line 4: This is working around [Array.Contains no longer works in latest .NET 9 SDK (9.0.200) (npgsql/efcore.pg#3461)](https://github.com/npgsql/efcore.pg/issues/3461#issuecomment-2670681690)

nuget.config
```diff
@@ -4,6 +4,7 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+    <add key="darc-pub-dotnet-aspire-2a8f48ea" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-dotnet-aspire-2a8f48ea/nuget/v3/index.json" />
```
eerhardt, line 7: nuget.config should be reverted once 9.1 is on nuget.org.

src/Catalog.API/Extensions/Extensions.cs
```diff
@@ -40,7 +41,7 @@ public static void AddApplicationServices(this IHostApplicationBuilder builder)
         if (builder.Configuration["OllamaEnabled"] is string ollamaEnabled && bool.Parse(ollamaEnabled))
         {
             builder.AddOllamaSharpEmbeddingGenerator("embedding");
-            builder.Services.AddEmbeddingGenerator(b => b.GetRequiredService<IEmbeddingGenerator<string, Embedding<float>>>())
+            builder.Services.AddEmbeddingGenerator(sp => (IEmbeddingGenerator<string, Embedding<float>>)sp.GetRequiredKeyedService<IOllamaApiClient>("embedding"))
```
eerhardt, line 44: FYI @stephentoub @SteveSandersonMS on the change here and in WebApp/Extensions/Extensions.cs.

These 2 things were pretty tricky.

1. I was getting infinite recursion on Ollama since `builder.Services.AddEmbeddingGenerator` is adding the `IEmbeddingGenerator<string, Embedding<float>>` service, and this was trying to get the same service. So it recurses forever.
2. In OpenAI code below, the call to `Build()` was failing because it was passing in an empty `IServiceProvider` to line 51, and that factory was trying to get the `OpenAIClient` out of the `sp`, which was empty.

captainsafia (approved):
LGTM!

Directory.Build.props
stephentoub: cc: @roji, @jaredpar

src/Catalog.API/Extensions/Extensions.cs
stephentoub: I didn't quite follow that. But is there something you'd recommend we change to make something easier?

src/WebApp/Extensions/Extensions.cs
```diff
@@ -101,7 +99,7 @@ private static void AddAIServices(this IHostApplicationBuilder builder)
         if (builder.Configuration["OllamaEnabled"] is string ollamaEnabled && bool.Parse(ollamaEnabled))
         {
             builder.AddOllamaSharpChatClient("chat");
-            builder.Services.AddChatClient(b => b.GetRequiredService<IChatClient>())
+            builder.Services.AddChatClient(sp => (IChatClient)sp.GetRequiredKeyedService<IOllamaApiClient>("chat"))
```
stephentoub, line 102: The Ollama component registers an IOllamaApiClient rather than an OllamaApiClient?

stephentoub (approved):
Thanks!

src/Catalog.API/Extensions/Extensions.cs
eerhardt: > But is there something you'd recommend we change to make something easier?

For the 2nd problem (the call to Build()), there is some awkwardness IMO. Take a look at the existing code:

https://github.com/dotnet/eShop/blob/1ba1cc7931296eea35932f4c04ef32d813f21a04/src/Catalog.API/Extensions/Extensions.cs#L50-L53

Code like this will never work.

The call to `Build()` doesn't pass an IServiceProvider. Thus the `sp` that is passed to the callback on line 50 will be empty, and won't contain any services. This is because the `Build()` method looks like this:

```C#
    public IEmbeddingGenerator<TInput, TEmbedding> Build(IServiceProvider? services = null)
    {
        services ??= EmptyServiceProvider.Instance;
```

Maybe `Build` should have a non-optional IServiceProvider parameter? There is a constructor that takes a `Func<IServiceProvider, IEmbeddingGenerator<TInput, TEmbedding>> innerGeneratorFactory` which won't work without a valid IServiceProvider being passed. Requiring the IServiceProvider would indicate to the caller that you shouldn't call `Build()` when adding services to your application, since you don't have an IServiceProvider yet.

For the infinite recursion problem, I'm not sure I have a suggestion. I think this was just confusion on my part trying to understand the old vs the new API usage. When writing new code, I don't think anyone would would implement the "innerGeneratorFactory" by directly returning the same service from the IServiceProvider . This is the function that is supposed to create the generator.

eerhardt: @eiriktsarpalis - FYI.

eiriktsarpalis: The call to `Build()` in this location is redundant, since it simply returns an `IChatClient` instance that is being discarded. If I'm honest I find the pattern used by the `AddChatClient`/`AddEmbeddingGenerator` methods somewhat strange and I can see why it could lead to confusion. Instead of doing
```C#
builder.Services.AddChatClient(sp => new Implementation())
     .UseOpenTelemetry() 
	 .UseLogging();
```
Why not just have the method return void and encourage the equivalent pattern:
```C#
builder.Services.AddChatClient(sp => new Implementation()
     .AsBuilder()
     .UseOpenTelemetry() 
	 .UseLogging()
     .Build());
```
It might be more lines of code but the intent is clearer.

eerhardt: Once [Return AspireOpenAIClientBuilder from AddOpenAIClientFromConfiguration (dotnet/aspire#7763)](https://github.com/dotnet/aspire/pull/7763) is merged, this code will get even more simple. (At least the OpenAI part will. We will need to update the Ollama Aspire integration to follow the same API pattern.)

```C#
builder.AddOpenAIClientFromConfiguration("openai");
builder.Services.AddEmbeddingGenerator(sp => sp.GetRequiredService<OpenAIClient>().AsEmbeddingGenerator(builder.Configuration["AI:OpenAI:EmbeddingModel"]!))
    .UseOpenTelemetry()
    .UseLogging();
```

becomes

```C#
builder.AddOpenAIClientFromConfiguration("openai")
    .AddEmbeddingGenerator(builder.Configuration["AI:OpenAI:EmbeddingModel"]!)
    .UseLogging();  // I'm not positive this is required, but I don't see Aspire calling it today
```

And with further changes like [Configure OpenAI models in the app host (dotnet/aspire#6577)](https://github.com/dotnet/aspire/pull/6577) we will be able to drop the call to `builder.Configuration`, and it will simply be:

```C#
builder.AddOpenAIClientFromConfiguration("openai-embedding")
    .AddEmbeddingGenerator()
    .UseLogging();  // I'm not positive this is required, but I don't see Aspire calling it today
```

Since the model/deployment name will be passed via the connection string named "openai-embedding".

stephentoub: > .UseLogging();  // I'm not positive this is required, but I don't see Aspire calling it today

Assuming UseOpenTelemetry is being used, UseLogging is probably not necessary.

eerhardt: > Assuming UseOpenTelemetry is being used

Yes, it is by default. The Aspire integration enables OpenTelemetry unless someone disables tracing:

https://github.com/dotnet/aspire/blob/3d3ccd4eb8b3f300a91b54f763cbad4ade71ad24/src/Components/Aspire.OpenAI/AspireOpenAIClientBuilderEmbeddingGeneratorExtensions.cs#L59-L61

```C#
        return builder.DisableTracing
            ? result
            : new OpenTelemetryEmbeddingGenerator<string, Embedding<float>>(result);
```

but diffing this code with what's in `UseOpenTelemetry()`, maybe we need to update it to pass an ILogger in:

https://github.com/dotnet/extensions/blob/576fa221f57f80423a4852e446cee2aa2e937aab/src/Libraries/Microsoft.Extensions.AI/Embeddings/OpenTelemetryEmbeddingGeneratorBuilderExtensions.cs#L36-L43

```C#
            loggerFactory ??= services.GetService<ILoggerFactory>();


            var generator = new OpenTelemetryEmbeddingGenerator<TInput, TEmbedding>(
                innerGenerator,
                loggerFactory?.CreateLogger(typeof(OpenTelemetryEmbeddingGenerator<TInput, TEmbedding>)),
                sourceName);
            configure?.Invoke(generator);
            return generator;
```

src/WebApp/Extensions/Extensions.cs
davidfowl: We should update this to use the new API @aaronpowell

eerhardt: Correct. That is pretty typical to register and use the interface (when available) instead of the concrete type. That way someone can easily mock it out in tests.

https://github.com/CommunityToolkit/Aspire/blob/3cda8176ea0f2026494df22514906dd4079880d7/src/CommunityToolkit.Aspire.OllamaSharp/AspireOllamaSharpExtensions.cs#L136-L148

src/Catalog.API/Extensions/Extensions.cs
eerhardt: > We will need to update the Ollama Aspire integration to follow the same API pattern.

This was done in https://github.com/CommunityToolkit/Aspire/pull/465. We can update here once it ships in a stable version of the Aspire Ollama integration.

cc @aaronpowell

src/WebApp/Extensions/Extensions.cs
eerhardt: > We should update this to use the new API @aaronpowell

We will. See https://github.com/dotnet/eShop/pull/759#discussion_r1970579654. We just need the new API to be stable.

PR #762, opened 2025-02-26
Update to CommunityToolkit Ollama 9.2

eerhardt:
Simplify the Ollama usage with the new API pattern.

The reason the WebApp.csproj's reference had to change is because before this change the `CommunityToolkit.Aspire.OllamaSharp` reference brought in `Aspire.Hosting` (yes into an app not the AppHost), which brought in `Google.Protobuf`. With the latest version, the `CT.Aspire.OllamaSharp` library no longer references `Aspire.Hosting`, which means no one brings `Google.Protobuf` and the WebApp doesn't build. While I was in there, I corrected which package it references. It wasn't using anything from the higher-level Grpc package. So I minimalized the dependency.

### DIFF
--- a/src/WebApp/Extensions/Extensions.cs
+++ b/src/WebApp/Extensions/Extensions.cs
@@ -98,11 +98,9 @@ public static class Extensions
     {
         if (builder.Configuration["OllamaEnabled"] is string ollamaEnabled && bool.Parse(ollamaEnabled))
         {
-            builder.AddOllamaSharpChatClient("chat");
-            builder.Services.AddChatClient(sp => (IChatClient)sp.GetRequiredKeyedService<IOllamaApiClient>("chat"))
-                .UseFunctionInvocation()
-                .UseOpenTelemetry(configure: t => t.EnableSensitiveData = true)
-                .UseLogging();
+            builder.AddOllamaApiClient("chat")
+                .AddChatClient()
+                .UseFunctionInvocation();
         }
         else
         {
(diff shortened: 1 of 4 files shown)


stephentoub (approved):

src/WebApp/Extensions/Extensions.cs
```diff
@@ -98,11 +98,9 @@ private static void AddAIServices(this IHostApplicationBuilder builder)
     {
         if (builder.Configuration["OllamaEnabled"] is string ollamaEnabled && bool.Parse(ollamaEnabled))
         {
-            builder.AddOllamaSharpChatClient("chat");
-            builder.Services.AddChatClient(sp => (IChatClient)sp.GetRequiredKeyedService<IOllamaApiClient>("chat"))
-                .UseFunctionInvocation()
-                .UseOpenTelemetry(configure: t => t.EnableSensitiveData = true)
-                .UseLogging();
+            builder.AddOllamaApiClient("chat")
```
aaronpowell, line 101: Since there's a functionality overlap between Ollama and OpenAI in which `UseFunctionInvocation` is added, should we capture the `ChatClientBuilder` (returned by `AddChatClient`) and then add the middleware outside of the `if` block so that we can illustrate how to have alternative AI providers but a consistent middleware pipeline across them?

eerhardt: Good idea.

That will be easier/possible once the OpenAI code can use [Return AspireOpenAIClientBuilder from AddOpenAIClientFromConfiguration (dotnet/aspire#7763)](https://github.com/dotnet/aspire/pull/7763). Since we need to manually add the `UseOpenTelemetry()` to the OpenAI pipeline right now, and it has to be after `UseFunctionInvocation()`, this wouldn't be as clean. But in the future it will be.

eerhardt: We had part of it in the 9.1 release. But we missed the "FromConfiguration" API when we first did the AIClientBuilder work. I didn't realize it until I came to eshop to update it to 9.1 that we missed that API.

PR #808, opened 2025-03-31
Re-enable audience validation

adityamandaleeka:
Not sure why this was disabled in the first place...

This pull request includes a small change to AuthenticationExtensions.cs to remove the line that disables audience validation in token validation parameters.

@halter73


danmoseley:
LGTM, but did ancient github history help?

PR #811, opened 2025-04-07
Updates to WinGet configuration files

mdanish-kh:
## Change

The new convention for WinGet configuration files is to place them in a `.config` directory with file extension `.winget`. With `.winget` file extension, users can directly launch the configuration from file explorer by double-clicking. See https://learn.microsoft.com/en-us/windows/package-manager/configuration/create#file-naming-convention

More updates:
- Remove `allowPrerelease: true` from Microsoft.WinGet.DSC as it's GA
- Add `securityContext: elevated` for resources that require elevation. The benefit here is that the configuration can be run in user mode, WinGet will ask for a single UAC at the start of configuration and run two separate processes (one elevated, and one with standard privileges). The resource that requires elevation will be run in the elevated process
- Fix some YAML indentation issues
- Use `Microsoft.VSCode.Dsc/VSCodeExtension` resource for installing extensions instead of the Script resource

### DIFF
--- a/.config/configuration.vs.winget
+++ b/.config/configuration.vs.winget
@@ -13,6 +13,8 @@ properties:
     - resource: Microsoft.Windows.Developer/DeveloperMode
       directives:
         description: Enable Developer Mode
+        # Requires elevation only for the set operation
+        securityContext: elevated
         allowPrerelease: true
       settings:
         Ensure: Present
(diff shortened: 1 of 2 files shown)


mdanish-kh:
@dotnet-policy-service agree

mdanish-kh:
@denelon - I know the name convention for the default configuration is `.config/configuration.winget` but here the default config may be based on user preference. One is meant for users that want to develop in VSCode & other is meant for Visual Studio. Not sure what the convention should be here. I went with `configuration.vs.winget` & `configuration.vsCode.winget` in this PR

claudiaregio:
@jamesmontemagno @captainsafia @adityamandaleeka can any of you review this PR?

danmoseley (approved):
seems reasonable as far as I can see. thanks!

PR #813, opened 2025-04-09
Update eshop to Aspire 9.2

joperezr:
FYI: @DamianEdwards @davidfowl @eerhardt


src/Catalog.API/Extensions/Extensions.cs
```diff
@@ -46,7 +46,7 @@ public static void AddApplicationServices(this IHostApplicationBuilder builder)
         else if (!string.IsNullOrWhiteSpace(builder.Configuration.GetConnectionString("openai")))
         {
             builder.AddOpenAIClientFromConfiguration("openai");
-            builder.Services.AddEmbeddingGenerator(sp => sp.GetRequiredService<OpenAIClient>().AsEmbeddingGenerator(builder.Configuration["AI:OpenAI:EmbeddingModel"]!))
+            builder.Services.AddEmbeddingGenerator(sp => sp.GetRequiredService<OpenAIClient>().GetEmbeddingClient(builder.Configuration["AI:OpenAI:EmbeddingModel"]!).AsIEmbeddingGenerator())
```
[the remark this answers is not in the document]
joperezr: It wouldn't be in aspire-docs as this is related to MEAI, but apart from that I don't think they necessarily document their breaking changes yet given they are in preview. AFAIK what they have been doing instead is that for each release they make containing breaking changes, they contribute PRs to different GH repositories to fix the breaks.

danmoseley (approved):

src/Catalog.API/Extensions/Extensions.cs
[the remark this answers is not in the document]
joperezr: We need to make the change here only because bumping aspire will transitively also bump MEAI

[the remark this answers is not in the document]
danmoseley: got it.

src/eShop.AppHost/Program.cs
```diff
@@ -99,6 +99,10 @@
            .WithEnvironment("WebhooksWebClient", webhooksClient.GetEndpoint(launchProfileName))
            .WithEnvironment("WebAppClient", webApp.GetEndpoint(launchProfileName));
 
+// Starting in Aspire 9.2, we can use the new DockerComposePublisher to generate a docker-compose file.
+// In order to do so, run 'dotnet run --publisher docker-compose --output-path ./docker-compose' to try it out.
+builder.AddDockerComposePublisher();
```
davidfowl, line 104: Did you test this?

joperezr: Yup, it generates the required files but when I tried doing docker compose up -d, things would run but couldn’t get https working correctly. I will look into this a bit more tomorrow.

src/Catalog.API/Extensions/Extensions.cs
davidfowl, line 49: This should be changed, we pass the model in the connection string and this should not be needed anymore.

cc @eerhardt

src/eShop.AppHost/Program.cs
DamianEdwards: That's expected right as once you're not running development the dev cert won't be loaded. I thought we made the docker publisher remove the https endpoints for this reason?

src/Catalog.API/Extensions/Extensions.cs
eerhardt: Fixed in the latest code.

Directory.Packages.props
```diff
@@ -38,20 +39,20 @@
     <PackageVersion Include="Microsoft.AspNetCore.Authentication.OpenIdConnect" Version="$(AspnetVersion)" />
     <PackageVersion Include="Microsoft.AspNetCore.Components.QuickGrid" Version="$(AspnetVersion)" />
     <PackageVersion Include="Microsoft.AspNetCore.Components.Web" Version="$(AspnetVersion)" />
-    <PackageVersion Include="Microsoft.AspNetCore.Identity.EntityFrameworkCore" Version="9.0.2" />
+    <PackageVersion Include="Microsoft.AspNetCore.Identity.EntityFrameworkCore" Version="9.0.4" />
```
eerhardt, line 42:
```suggestion
    <PackageVersion Include="Microsoft.AspNetCore.Identity.EntityFrameworkCore" Version="$(AspnetVersion)" />
```

joperezr:
Thanks folks!

PR #834, opened 2025-05-09
Use Microsoft.Windows.Settings module for enabling developer mode

mdanish-kh:
Using the [Microsoft.Windows.Settings](https://www.powershellgallery.com/packages/Microsoft.Windows.Settings/) module is the now the recommended way of configuring Windows Settings, including developer mode.


denelon:
Thanks @mdanish-kh

PR #890, opened 2025-09-04
Upgrading eshop to .NET 10 and Aspire 9.4.2

joperezr:


src/eShop.ServiceDefaults/OpenApiOptionsExtensions.cs
```diff
@@ -116,19 +116,17 @@ public static OpenApiOptions ApplyAuthorizationChecks(this OpenApiOptions option
                 return Task.CompletedTask;
             }
 
+            operation.Responses ??= new OpenApiResponses();
```
joperezr, line 119: cc: @captainsafia can you help me validate these changes as part of our migration from net9 to net10? These were all Copilot-made changes in response to some breaking changes that happened in openapi, so would be good to confirm these look good.

.aspire/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "appHostPath": "../src/eShop.AppHost/eShop.AppHost.csproj"
+}
```
joperezr, line 3: Assuming we don't want to checkin this file as it is likely just the result of running `aspire run`. If @captainsafia confirms I can remove.

src/eShop.ServiceDefaults/OpenApiOptionsExtensions.cs
danmoseley: I don't think I see one? @captainsafia

Directory.Packages.props
```diff
@@ -2,12 +2,12 @@
   <PropertyGroup>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
-    <AspnetVersion>9.0.7</AspnetVersion>
-    <MicrosoftExtensionsVersion>9.0.7</MicrosoftExtensionsVersion>
-    <AspireVersion>9.4.0</AspireVersion>
-    <AspireUnstablePackagesVersion>9.4.0-preview.1.25378.8</AspireUnstablePackagesVersion>
+    <AspnetVersion>10.0.0-preview.7.25380.108</AspnetVersion>
```
danmoseley, line 5: rc1 for these now I guess

src/HybridApp/Services/CatalogService.cs
```diff
@@ -13,41 +13,41 @@ public class CatalogService(HttpClient httpClient) : ICatalogService
     public Task<CatalogItem?> GetCatalogItem(int id)
     {
         var uri = $"{remoteServiceBaseUrl}items/{id}?api-version=2.0";
-        return httpClient.GetFromJsonAsync<CatalogItem>(uri);
+        return httpClient.GetFromJsonAsync(uri, CatalogJsonContext.Default.CatalogItem);
```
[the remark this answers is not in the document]
joperezr: You don't need to do this to migrate. This was to address some new trimmer warnings. Basically in .NET 10 new APIs were annotated for trimming warnings, and the way we had the code before made it so the trimmer wouldn't know what needs to be kept, so warnings were logged. This change makes it so that the trimmer understands which types need to be kept, so it fixes the warnings.

src/Catalog.API/Catalog.API_v2.json
```diff
@@ -241,6 +248,7 @@
             "description": "The catalog item id",
             "required": true,
             "schema": {
+              "pattern": "^-?(?:0|[1-9]\\d*)$",
```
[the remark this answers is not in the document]
joperezr: mm these changes were done by Copilot automatically. How can I regenerate this file to make sure it looks right?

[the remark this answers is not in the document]
joperezr: looks like I had pending changes for this, I just pushed the latest generated files based on the latest build

PR #903, opened 2025-10-12
Fix paths for Bootstrap and jQuery in layout files

Seesi:
The layout references bootstrap and jquery files within `dist` subdirectories, but the actual folder structure in `wwwroot/lib` was missing these `dist` folders.
This caused 404 errors and left the login page unstyled.


Seesi:
@dotnet-policy-service agree

PR #912, opened 2025-10-30
Update in README eShop version from .NET 9 to .NET 10

TachyonTracer:
What I changed:
- Replaced reference to .NET 9 in the README with .NET 10.

Notes for reviewers:
- This is a documentation-only change.
- I did a quick search for other ".NET 9" references — please double-check CI configs and global.json; if you want I can update them in the same PR (or open a follow-up PR).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,13 @@ A reference .NET application implementing an e-commerce website using a services
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


allexgut:
It would also be right to also update global.json, no?

PR #980, opened 2026-04-19
Update to Asp.Versioning to v10-preview2

mikekistler:
This PR updates the Asp.Versioning package to 10.0.0-preview.2, which provides a cleaner and simpler versioning configuration for .NET 10 projects.

### DIFF
--- a/.spectral.yml
+++ b/.spectral.yml
@@ -1,6 +1,7 @@
 extends: spectral:oas
 rules:
   info-contact: off
+  oas3-api-servers: off
 
   success-response:
     description: All operations should have a success response.
--- a/src/Catalog.API/Catalog.API.http
+++ b/src/Catalog.API/Catalog.API.http
@@ -1,31 +1,44 @@
-﻿@Catalog.API_HostAddress = http://localhost:5222
+﻿@HostAddress = http://localhost:5222
 @ApiVersion = 1.0
 
-GET {{Catalog.API_HostAddress}}/openapi/v1.json
+# Scalar: http://localhost:5222/scalar/v1
+GET {{HostAddress}}/openapi/v1.json
 
 ###
 
-GET {{Catalog.API_HostAddress}}/api/catalog/items?api-version={{ApiVersion}}
+GET {{HostAddress}}/openapi/v2.json
 
 ###
 
-GET {{Catalog.API_HostAddress}}/api/catalog/items/type/1/brand/2?api-version={{ApiVersion}}
+GET {{HostAddress}}/api/catalog/items
+
+###
+
+GET {{HostAddress}}/api/catalog/items?api-version=1
+
+###
+
+GET {{HostAddress}}/api/catalog/items?api-version={{ApiVersion}}
+
+###
+
+GET {{HostAddress}}/api/catalog/items/type/1/brand/2?api-version={{ApiVersion}}
 
 ###
 
 # A request with an unknown API version returns a 400 ProblemDetails response
 
-GET {{Catalog.API_HostAddress}}/api/catalog/items/463/pic?api-version=99
+GET {{HostAddress}}/api/catalog/items/463/pic?api-version=99
 
 ###
 
 # A request with an unknown item id returns a 404 NotFound with empty response body
 
-GET {{Catalog.API_HostAddress}}/api/catalog/items/463/pic?api-version={{ApiVersion}}
+GET {{HostAddress}}/api/catalog/items/463/pic?api-version={{ApiVersion}}
 
 ###
 
-PUT {{Catalog.API_HostAddress}}/api/catalog/items?api-version={{ApiVersion}}
+PUT {{HostAddress}}/api/catalog/items?api-version={{ApiVersion}}
 content-type: application/json
 
 {
--- a/src/Catalog.API/Catalog.API.http
+++ b/src/Catalog.API/Catalog.API.http
@@ -1,7 +1,6 @@
 ﻿@HostAddress = http://localhost:5222
 @ApiVersion = 1.0
 
-# Scalar: http://localhost:5222/scalar/v1
 GET {{HostAddress}}/openapi/v1.json
 
 ###
@@ -10,11 +9,13 @@ GET {{HostAddress}}/openapi/v2.json
 
 ###
 
-GET {{HostAddress}}/api/catalog/items
+# Scalar: http://localhost:5222/scalar/v1
 
 ###
 
-GET {{HostAddress}}/api/catalog/items?api-version=1
+# api-version is required, so this request will fail
+
+GET {{HostAddress}}/api/catalog/items
 
 ###
 
--- a/src/eShop.ServiceDefaults/OpenApi.Extensions.cs
+++ b/src/eShop.ServiceDefaults/OpenApi.Extensions.cs
@@ -21,16 +21,24 @@ public static partial class Extensions
             return app;
         }
 
-        app.MapOpenApi();
+        app.MapOpenApi().WithDocumentPerVersion();
 
         if (app.Environment.IsDevelopment())
         {
+            var descriptions = app.DescribeApiVersions();
+            var defaultDocument = descriptions.Count > 0 ? descriptions[^1].GroupName : "v1";
+
             app.MapScalarApiReference(options =>
             {
                 // Disable default fonts to avoid download unnecessary fonts
                 options.DefaultFonts = false;
+
+                foreach (var description in descriptions)
+                {
+                    options.AddDocument(description.GroupName, description.GroupName, isDefault: description.GroupName == defaultDocument);
+                }
             });
-            app.MapGet("/", () => Results.Redirect("/scalar/v1")).ExcludeFromDescription();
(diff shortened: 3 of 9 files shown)


src/eShop.ServiceDefaults/OpenApi.Extensions.cs
```diff
@@ -21,16 +21,24 @@ public static IApplicationBuilder UseDefaultOpenApi(this WebApplication app)
             return app;
         }
 
-        app.MapOpenApi();
+        app.MapOpenApi().WithDocumentPerVersion();
 
         if (app.Environment.IsDevelopment())
         {
+            var descriptions = app.DescribeApiVersions();
+            var defaultDocument = descriptions.Count > 0 ? descriptions[^1].GroupName : "v1";
```
halter73, line 29: Nice use of `^1`! Is there a particular logic to defaulting to the last one? I presume their normally listed oldest to newest? Is this documented anyewhere?

Edit: I can't find the documentation, but this does appear to be the behavior.

src/eShop.ServiceDefaults/OpenApi.Extensions.cs
```diff
@@ -57,19 +65,21 @@ public static IHostApplicationBuilder AddDefaultOpenApi(
         {
             // the default format will just be ApiVersion.ToString(); for example, 1.0.
             // this will format the version as "'v'major[.minor][-status]"
-            var versioned = apiVersioning.AddApiExplorer(options => options.GroupNameFormat = "'v'VVV");
-            string[] versions = ["v1", "v2"];
```
halter73, line 61: This was kind of nice because it was self-documenting. I can see the argument for removing it if it could potentially get out of sync though.

src/eShop.ServiceDefaults/OpenApiOptionsExtensions.cs
```diff
@@ -139,8 +139,10 @@ public static OpenApiOptions ApplyOperationDeprecatedStatus(this OpenApiOptions
     {
         options.AddOperationTransformer((operation, context, cancellationToken) =>
         {
-            var apiDescription = context.Description;
-            operation.Deprecated |= apiDescription.IsDeprecated();
+            operation.Deprecated |= context.Description.ActionDescriptor.EndpointMetadata
+                .OfType<ObsoleteAttribute>()
+                .Any();
```
halter73: Nit: Did the `IsDeprecated()` API get removed? Either way, it might be better to gate this with an if. I don't think `||=` exists, and `|=` doesn't short-circuit since it's a bitwise-or.


```suggestion
            if (!operation.Deprecated &&
                context.Description.ActionDescriptor.EndpointMetadata.OfType<ObsoleteAttribute>().Any())
            {
                operation.Deprecated = true;
            }
```

`IsDeprecated()` reads a lot nicer though.

src/eShop.ServiceDefaults/OpenApiOptionsExtensions.cs
```diff
@@ -154,17 +156,13 @@ public static OpenApiOptions ApplyApiVersionDescription(this OpenApiOptions opti
             var apiVersionParameter = operation.Parameters?.FirstOrDefault(p => p.Name == "api-version");
             if (apiVersionParameter is not null)
             {
-                apiVersionParameter.Description = "The API version, in the format 'major.minor'.";
+                // Fix up the API version parameter description and example
+                // apiVersionParameter.Description = "The API version, in the format 'major.minor'.";
```
halter73: Did you mean to comment this out? Should we just delete the line?

```suggestion
```

src/eShop.ServiceDefaults/OpenApiOptionsExtensions.cs
```diff
@@ -154,17 +156,13 @@ public static OpenApiOptions ApplyApiVersionDescription(this OpenApiOptions opti
...
                 {
-                    switch (context.DocumentName) {
-                        case "v1":
-                            targetSchema.Example = JsonNode.Parse("\"1.0\"");
-                            break;
-                        case "v2":
-                            targetSchema.Example = JsonNode.Parse("\"2.0\"");
-                            break;
-                    }
+                    // Set the example to the default value and then clear the default
```
halter73: Why do we need to "clear the default"? Not saying we don't. It's just not that obvious to me.

.spectral.yml
```diff
@@ -1,6 +1,7 @@
 extends: spectral:oas
 rules:
   info-contact: off
+  oas3-api-servers: off
```
halter73, line 4: This looks like some sort of suppression. Why do we need it?

https://docs.stoplight.io/docs/spectral/4dec24461f3af-open-api-rules#oas3-api-servers

halter73 (approved):

.spectral.yml
halter73: I asked AI about this:

> With MapOpenApi().WithDocumentPerVersion(), the document is generated statically (no incoming-request
context), so ASP.NET Core's OpenAPI no longer emits a servers array. Spectral's oas3-api-servers rule
(inherited from spectral:oas) requires one, so without the suppression every generated openapi/v*.json
would warn. (Spectral isn't wired into CI here — it's editor/local only — but still noisy.) That's almost
certainly why Mike added it

It suggested this as a possible alternative though:

> Yes — better alternatives exist. The cleanest is to add the servers entry yourself via an OpenAPI document transformer, which is the actual fix rather than silencing the rule. A few options, roughly best → worst:
> 1. Add a document transformer that emits servers (recommended) In OpenApiOptionsExtensions.cs, alongside
the other Apply* helpers:
>
> ```csharp
> public static OpenApiOptions ApplyServers(this OpenApiOptions options, IConfiguration config)
>  {
>      options.AddDocumentTransformer((document, context, ct) =>
>      {
>          document.Servers = [new OpenApiServer { Url = config["OpenApi:ServerUrl"] ?? "/" }];
>          return Task.CompletedTask;
>      });
>      return options;
>  }
> ```
> 2. Scoped Spectral override (.spectral.yml overrides: section) — disable just for the generated
> openapi/*.json paths instead of globally. Better than a blanket off, but still papering over the issue.
> 
> 3. Lower severity to hint rather than off — keeps it visible without failing.
> 
> Why option 1 is strictly better: a servers entry isn't just lint candy — clients/codegen tools (Kiota,
> NSwag, openapi-generator) use it to compute base URLs. Without it, generated clients default to whatever
> the consumer hardcodes. Since this repo already runs Scalar UI and presumably wants the OpenAPI docs to
> be useful artifacts, populating servers is the right move.

src/Catalog.API/Catalog.API.http
```diff
@@ -1,31 +1,44 @@
...
 
 ###
 
-GET {{Catalog.API_HostAddress}}/api/catalog/items?api-version={{ApiVersion}}
+GET {{HostAddress}}/openapi/v2.json
 
 ###
 
-GET {{Catalog.API_HostAddress}}/api/catalog/items/type/1/brand/2?api-version={{ApiVersion}}
+GET {{HostAddress}}/api/catalog/items
```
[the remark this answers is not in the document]
mikekistler: This is intentional, to demonstrate the error response when api-version is omitted. I'll add a comment for this.

src/eShop.ServiceDefaults/OpenApi.Extensions.cs
mikekistler: I think that's the point -- only define these once to avoid errors from partial updates.

src/eShop.ServiceDefaults/OpenApi.Extensions.cs
```diff
@@ -57,19 +65,21 @@ public static IHostApplicationBuilder AddDefaultOpenApi(
...
-            foreach (var description in versions)
-            {
-                builder.Services.AddOpenApi(description, options =>
+            apiVersioning.AddApiExplorer(options =>
+                {
+                    options.GroupNameFormat = "'v'VVV";
+                    options.DefaultApiVersionParameterDescription = "The API version, in the format 'major.minor'.";
```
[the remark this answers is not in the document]
mikekistler: I'm actually hoping to correct that behavior.

src/eShop.ServiceDefaults/OpenApiOptionsExtensions.cs
mikekistler: I think `isDeprecated()` was an extension method added in v8 of Asp.Versioning, and it was dropped in v10 because v10 adds `VersionedApiDescription` with all the versioning extensions.

Good catch to not override `operation.Deprecated`. I fixed this slightly differently.

src/eShop.ServiceDefaults/OpenApiOptionsExtensions.cs
mikekistler: Because the parameter is required, so having a default is confusing (IMHO).

.spectral.yml
mikekistler: That's a surprisingly good AI response. The transformer would be a good addition but I think the one recommended is too simplistic, so I'm going to defer this to later work.

PR #985, opened 2026-04-30
Update to Aspire 13.3

eerhardt:
Also bump versions of other packages to latest.

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -4,19 +4,13 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="nuget" value="https://api.nuget.org/v3/index.json" />
-    <add key="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-microsoft-aspire-be8c19e4/nuget/v3/index.json" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-microsoft-aspire-be8c19e4/nuget/v3/index.json" />
   </packageSources>
+
   <packageSourceMapping>
     <!-- key value for <packageSource> should match key values from <packageSources> element -->
     <clear />
     <packageSource key="nuget">
       <package pattern="*" />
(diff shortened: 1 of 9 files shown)


nuget.config
```diff
@@ -4,13 +4,19 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+    <add key="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-microsoft-aspire-be8c19e4/nuget/v3/index.json" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/darc-pub-microsoft-aspire-be8c19e4/nuget/v3/index.json" />
```
eerhardt: This will be reverted once 13.3 officially ships.

src/Catalog.API/Catalog.API.json
```diff
@@ -38,8 +38,10 @@
             "description": "The API version, in the format 'major.minor'.",
             "required": true,
             "schema": {
-              "type": "string",
-              "example": "1.0"
+              "enum": [
```
eerhardt: @mikekistler @halter73 - I updated `Asp.Versioning.*` packages. Are these changes expected?

mikekistler: It looks like this is a recent change in the latest Asp.Versioning package:

https://github.com/dotnet/aspnet-api-versioning/releases/tag/v10.0.0

> OpenAPI documents use <code>"enum": ["1.0"]</code> instead of <code>"default": "1.0"</code>

We have a transformer that repaced `default` with `example`. We should just update that to remove `enum` and add `example`.

eerhardt: Ok, I reverted the upgrade and opened https://github.com/dotnet/eShop/issues/989 for this.

IEvangelist (approved):

PR #988, opened 2026-05-09
Fix inaccurate comment in SaveEntitiesAsync

rhodi2436:
The original comment claimed that dispatching domain events AFTER SaveChanges (option B) would result in multiple transactions. This was true before TransactionBehavior was introduced, but since TransactionBehavior wraps the entire command handler in an explicit transaction, both orderings now result in a single database transaction.

Update the comment to accurately describe the real trade-off between the two approaches, and note that the explicit transaction uses ReadCommitted isolation (see BeginTransactionAsync), which means handlers in option B can observe already-flushed state via SQL queries within the same transaction.


rhodi2436:
@dotnet-policy-service agree

rhodi2436 (commented):
Done

PR #1001, opened 2026-07-11
Fix AddNewOrder test validation and integration event discovery

### LINKED ISSUE #943: AddNewOrder in Ordering.FunctionalTests is not properly tested
`Shop.Ordering.FunctionalTests.OrderingApiTests.AddNewOrder` in the `OrderingFunctionalTest` project cannot be considered properly tested for the following reasons.

### 1. Parameters that cause validation errors are being passed

The `POST api/orders` API receives a `CreateOrderRequest`.  
An instance of this class is converted into `IdentifiedCommand<CreateOrderCommand>` and passed to `mediator.Send`.

However, due to `eShop.Ordering.API.Application.Validations.CreateOrderCommandValidator`, a validation error occurs when the command is sent.

Because of the following code in `IdentifiedCommandHandler`, validation errors for `CreateOrderCommand` are caught by exception handling, and the return value of the `Handle` method becomes `false` (since the method returns `default` for a `bool` return type).

https://github.com/dotnet/eShop/blob/b832741187b7a1d6c993d5a0321a622d31e75203/src/Ordering.API/Application/Commands/IdentifiedCommandHandler.cs#L87

```cs
    var result = await _mediator.Send(command, cancellationToken);

    _logger.LogInformation(
        "Command result: {@Result} - {CommandName} - {IdProperty}: {CommandId} ({@Command})",
        result,
        commandName,
        idProperty,
        commandId,
        command);

    return result;
}
catch
{
    return default;
}
```

As a result, even when `mediator.Send` returns `false`, the `POST api/orders` API still returns `200 OK`, causing the test to pass incorrectly.

```cs
var result = await services.Mediator.Send(requestCreateOrder);

if (result)
{
    services.Logger.LogInformation("CreateOrderCommand succeeded - RequestId: {RequestId}", requestId);
}
else
{
    services.Logger.LogWarning("CreateOrderCommand failed - RequestId: {RequestId}", requestId);
}

return TypedResults.Ok();
```

---

### 2. A `Microsoft.EntityFrameworkCore.DbUpdateException` occurs because `cardExpirationDate.Kind` is `Kind.Local`  

(This issue may be environment-dependent.)

If issue 1 is fixed so that validation succeeds, `AddNewOrder` fails due to an exception.

```cs
var cardExpirationDate = Convert.ToDateTime("2123-12-22T12:34:24.334Z");
var OrderRequest = new CreateOrderRequest(
    UserId: "1",
    UserName: "TestUser",
    City: "test city",
    Street: "test street",
    State: "test state",
    Country: "test country",
    ZipCode: "00001",
    CardNumber: "XXXXXXXXXXXX0005",
    CardHolderName: "Test User",
    CardExpiration: cardExpirationDate,
    CardSecurityNumber: "000",
    CardTypeId: 1,
    Buyer: "TestBuyer",
    Items: new List<BasketItem> { item });
```

When writing `DateTime` values to PostgreSQL, the `Kind` must be `Kind.Utc`.  
However, since `cardExpirationDate` is initialized as shown above, its `Kind` becomes `Kind.Local`, which is presumed to be the cause of the exception.

If the code is changed as follows, the exception does not occur:

```cs
var cardExpirationDate = DateTime.Parse(
    "2123-12-22T12:34:24Z",
    null,
    System.Globalization.DateTimeStyles.AdjustToUniversal);
```

---

### 3. `IntegrationEventLogService` fails to deserialize `EventLogEntry` when publishing `OrderStartedIntegrationEvent`

After fixing issue 2, another issue arises.  
This may be better handled as a separate issue.

Because derived classes of `IntegrationEvent` are not included in the FunctionalTest assembly, the following code in  
`eshop.IntegrationEventLogEF.Services.IntegrationEventLogService<TContext>` results in an empty array:

https://github.com/dotnet/eShop/blob/b832741187b7a1d6c993d5a0321a622d31e75203/src/IntegrationEventLogEF/Services/IntegrationEventLogService.cs#L13

````cs
// In FunctionalTests, the Entry Assembly is the DLL built from the test project.
// Derived IntegrationEvent classes do not exist there.
_eventTypes = Assembly.Load(Assembly.GetEntryAssembly().FullName)
    .GetTypes()
    .Where(t => t.Name.EndsWith(nameof(IntegrationEvent)))
    .ToArray();
````

Since `_eventTypes` is an empty array, `RetrieveEventLogsPendingToPublishAsync` in the same class fails at the following code:

https://github.com/dotnet/eShop/blob/b832741187b7a1d6c993d5a0321a622d31e75203/src/IntegrationEventLogEF/Services/IntegrationEventLogService.cs#L28

````cs
// Since _eventTypes is empty, a NullReferenceException occurs
return result.OrderBy(o => o.CreationTime)
    .Select(e => e.DeserializeJsonContent(
        _eventTypes.FirstOrDefault(t => t.Name == e.EventTypeShortName)));
````

I checked other tests but could not find any others that publish `IntegrationEvent`s.  
Therefore, it is fair to say that tests involving the publication of integration events are not properly covered by FunctionalTests.

onurnafiguzel:
Fixes #943

## Problem

The `AddNewOrder` functional test in `Ordering.FunctionalTests` passed even though the
order was never actually created. Issue #943 identified three underlying problems:

1. **Validation failures were masked.** `CreateOrderAsync` always returned `200 OK`, even
   when the command failed, so an invalid request looked successful.
2. **`DateTime.Kind` mismatch.** The test built the card expiration date with
   `Convert.ToDateTime(...)`, producing `Kind.Local`, which Npgsql rejects for
   `timestamp with time zone` columns.
3. **Integration event types could not be resolved under tests.**
   `IntegrationEventLogService` loaded event types from `Assembly.GetEntryAssembly()`.
   In functional tests the entry assembly is the test host, which does not contain the
   derived `IntegrationEvent` types, so publishing an order's integration event threw and
   returned `500`.

## Changes

- **`OrdersApi.CreateOrderAsync`**: now returns `BadRequest` when the command fails,
  consistent with the `Cancel`/`Ship` endpoints, instead of always returning `Ok`.
- **`IntegrationEventLogService`**: the assembly to scan for event types is now injected
  via the constructor instead of relying on the entry assembly.
- **`Ordering.API` / `Catalog.API` DI registration**: each API now passes its own
  assembly (`typeof(Extensions).Assembly`) when registering the log service.
- **`OrderingApiTests`**:
  - Renamed the existing invalid-data test to `AddNewOrderWithInvalidDataFails`
    (asserts `BadRequest`).
  - Added a new `AddNewOrder` test using valid data (future UTC expiration date, complete
    address, valid security number) that asserts `OK`.

## Testing

- `Ordering.FunctionalTests`: 12/12 passing
- `Ordering.UnitTests`: 41/41 passing

### DIFF
--- a/src/Catalog.API/Extensions/Extensions.cs
+++ b/src/Catalog.API/Extensions/Extensions.cs
@@ -24,7 +24,11 @@ public static class Extensions
         builder.Services.AddMigration<CatalogContext, CatalogContextSeed>();
 
         // Add the integration services that consume the DbContext
-        builder.Services.AddTransient<IIntegrationEventLogService, IntegrationEventLogService<CatalogContext>>();
+        // Pass this project's assembly so the integration event types can be resolved for deserialization. Using the entry assembly breaks under functional tests, where the entry assembly is the test host and does not contain the derived IntegrationEvent types.
+        builder.Services.AddTransient<IIntegrationEventLogService>(sp =>
+            new IntegrationEventLogService<CatalogContext>(
+                sp.GetRequiredService<CatalogContext>(),
+                typeof(Extensions).Assembly));
 
         builder.Services.AddTransient<ICatalogIntegrationEventService, CatalogIntegrationEventService>();
 
--- a/src/IntegrationEventLogEF/Services/IntegrationEventLogService.cs
+++ b/src/IntegrationEventLogEF/Services/IntegrationEventLogService.cs
@@ -7,10 +7,10 @@ public class IntegrationEventLogService<TContext> : IIntegrationEventLogService,
     private readonly TContext _context;
     private readonly Type[] _eventTypes;
 
-    public IntegrationEventLogService(TContext context)
+    public IntegrationEventLogService(TContext context, Assembly eventTypesAssembly)
     {
         _context = context;
-        _eventTypes = Assembly.Load(Assembly.GetEntryAssembly().FullName)
+        _eventTypes = eventTypesAssembly
             .GetTypes()
             .Where(t => t.Name.EndsWith(nameof(IntegrationEvent)))
             .ToArray();
--- a/src/Ordering.API/Apis/OrdersApi.cs
+++ b/src/Ordering.API/Apis/OrdersApi.cs
@@ -157,13 +157,11 @@ public static class OrdersApi
             if (result)
             {
                 services.Logger.LogInformation("CreateOrderCommand succeeded - RequestId: {RequestId}", requestId);
-            }
-            else
-            {
-                services.Logger.LogWarning("CreateOrderCommand failed - RequestId: {RequestId}", requestId);
+                return TypedResults.Ok();
             }
 
-            return TypedResults.Ok();
+            services.Logger.LogWarning("CreateOrderCommand failed - RequestId: {RequestId}", requestId);
+            return TypedResults.BadRequest("Create order failed to process.");
         }
     }
 }
--- a/src/Ordering.API/Extensions/Extensions.cs
+++ b/src/Ordering.API/Extensions/Extensions.cs
@@ -21,7 +21,11 @@ internal static class Extensions
         services.AddMigration<OrderingContext, OrderingContextSeed>();
 
         // Add the integration services that consume the DbContext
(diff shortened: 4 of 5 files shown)


onurnafiguzel:
@dotnet-policy-service agree

